PR #14, opened 2021-11-03
[Types] Add new entity type and action

jainsneha23:
Add new entity type "search", and action "create"


packages/nova-types/src/nova-commanding-centralized.interface.ts
```diff
@@ -11,7 +11,7 @@
  * expressed within the larger system).
  */
 
-export interface NovaCentralizedCommanding {
+ export interface NovaCentralizedCommanding {
```
kerrynf, line 14: Can you revert this additional space? :)

kerrynf (approved):

PR #16, opened 2021-12-02
.mjs fix for taggedNode

kenotron:
## Problem

`require` is not legal in ES modules. We can only use the `import` keyword

## What is changing in this PR

1. Replacing the "require" with a static import:
    ```js
    /**
     * Imports graphql-js-tag from a peer dependency.
     * 
     * @graphitation/graphql-js-tag is expected to be provided by the consumer using this library.
     * Failure to do so will result in a TypeScript compilation or a bundling error - there is no
     * need to incur runtime costs to run checks for this library as a dependency.
     */
    import { graphql as graphqlJsTag } from "@graphitation/graphql-js-tag";
    ```
2. Moves the `@graphitation/graphql-js-tag` as peerDep + devDep
3. Adds a little note in README.md about the pre-req
4. Removed the use of "invariant" to do checks for the dep

### DIFF
--- a/packages/nova-react/README.md
+++ b/packages/nova-react/README.md
@@ -3,3 +3,7 @@
 The Nova Facade is a set of interfaces that represent the core framework dependencies of a data-backed UI component in a large application. This allows high value components to be written in a host agnostic fashion and used within any host that implements the Nova contracts.
 
 This package provides a React specific implementation of the contexts etc required to make the host implementations accessible to the Nova component code, with the sister package @nova/types providing a framework agnostic set of contracts.
+
+## Prerequisites
+
+In order to use this package, make sure to add `@graphitation/graphql-js-tag` as a dependency in your `package.json` along side with `@nova/react`.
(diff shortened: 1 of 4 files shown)


packages/nova-react/README.md
```diff
@@ -3,3 +3,7 @@
 The Nova Facade is a set of interfaces that represent the core framework dependencies of a data-backed UI component in a large application. This allows high value components to be written in a host agnostic fashion and used within any host that implements the Nova contracts.
 
 This package provides a React specific implementation of the contexts etc required to make the host implementations accessible to the Nova component code, with the sister package @nova/types providing a framework agnostic set of contracts.
+
+## Prerequisites
+
+In order to use this package, make sure to add `@graphitation/graphql-js-tag` as a dependency in your `package.json` along side with `@nova/react`.
```
alloy, line 9:
```suggestion
In order to use this package without compilation of documents, i.e. to parse string documents at runtime, make sure to add `@graphitation/graphql-js-tag` as a dependency in your `package.json` along side with `@nova/react`.
```

alloy (approved):
Minor wording change, but otherwise LGTM 👍

PR #29, opened 2022-02-02
[monorepo] Respect index files when building mjs modules

alloy:
Equivalent of https://github.com/microsoft/graphitation/pull/115


kerrynf:
Ping on this - should we merge this in, @freiksenet @alloy?

PR #33, opened 2022-03-04
Nova Eventing

kerrynf:
Nova Eventing is designed to replace Nova Centralized Commanding as the preferred method of communication from Nova UI components.

Unlike commanding, which couples the UI components to the knowledge of who they are communicating with, eventing allows the components to define their own event shapes, which handlers in the higher business logic layers can interpret and respond to as needed.

### DIFF
--- a/packages/nova-react/src/eventing/react-event-source-mapper.ts
+++ b/packages/nova-react/src/eventing/react-event-source-mapper.ts
@@ -1,0 +1,85 @@
+import { EventWrapper, InputType, Source } from "@nova/types";
+import { ReactEventWrapper } from "./nova-eventing-provider";
+
+/**
+ * Maps DOM event type strings to the Nova InputType.
+ *
+ * Initial implementation is not exhaustive and should be
+ * extended as new requirements arise.
+ *
+ * Note following excluded event types that could be considered:
+ * scroll, input, submit, toggle
+ */
+const typeMap = {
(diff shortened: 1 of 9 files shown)


packages/nova-react/src/eventing/react-event-source-mapper.ts
```diff
@@ -0,0 +1,85 @@
...
+  keyup: InputType.keyboard,
+  keypress: InputType.keyboard,
+  mousedown: InputType.mouse,
+  mouseenter: InputType.mouse,
+  mouseleave: InputType.mouse,
+  mousemove: InputType.mouse,
+  mouseover: InputType.mouse,
+  mouseout: InputType.mouse,
+  mouseup: InputType.mouse,
+  touchcancel: InputType.touch,
```
jurokapsiar, line 25: it would be good to understand if pen events manifest as touch as well. But I don't have a pen device :(

jurokapsiar:
created a testing codesandbox: https://codesandbox.io/s/condescending-blackwell-8ln0ur?file=/index.tsx

jurokapsiar (approved):

packages/nova-react/src/eventing/react-event-source-mapper.ts
petr-duda: Tried it on Surface laptop and pen event is recognized correctly.
image

PR #34, opened 2022-03-08
Upgrade `graphitation` deps

sergey-stoyan:
**Problem:**
There is memory leak in Nova components in TMP atm for the new Apollo client 3.5.x

**Solution:**
Root cause is in `graphql-js-tag` (the fix is already applied and released in 0.9.0), that is used by `nova-facade`.
The goal of this PR is to upgrade dependencies so that correct version of `graphql-js-tag` is installed.


freiksenet (approved):

alloy:
@sergey-stoyan Was there an actual leak in our tag package? Or are you referring to the bloated AST with location data and/or Apollo Client not being able to normalise queries due to location data in AST?

sergey-stoyan:
> @sergey-stoyan Was there an actual leak in our tag package? Or are you referring to the bloated AST with location data and/or Apollo Client not being able to normalise queries due to location data in AST?

That's what we were discussed - new Apollo Client not being able to process queries with locations in AST

PR #63, opened 2022-12-10
Fix Source timeStamp to consistently use Epoch time

kerrynf:
React SyntheticEvent timestamps are based on the DOM event timeStamps that are derived as 'time since document load'. (https://developer.mozilla.org/en-US/docs/Web/API/Event/timeStamp).

The timeStamp field on the Nova Source object is expected to use the Javascript Epoch time format (milliseconds since Jan 1 1970).

This change uses the underlying event data attached to the React SyntheticEvent to convert the UIEvent timeStamp into standard Epoch format, and falls back to Date.now() if the data is not available.

### DIFF
--- a/change/@nova-react-30a04f5f-7bd3-4681-8961-7b6e6de5a434.json
+++ b/change/@nova-react-30a04f5f-7bd3-4681-8961-7b6e6de5a434.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Fix Source timestamp to use consistent Epoch time",
+  "packageName": "@nova/react",
+  "email": "kerryn@zyridium.net",
+  "dependentChangeType": "patch"
+}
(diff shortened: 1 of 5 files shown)


alloy (approved):
This looks good, but I just want to note that this wasn't always the case (before 2016), but probably not a case we need to consider?

See e.g. this polyfill https://github.com/majido/high-resolution-timestamp-polyfill/blob/master/translate-timeStamp.js

kerrynf:
> This looks good, but I just want to note that this wasn't always the case (before 2016), but probably not a case we need to consider?
> 
> See e.g. this polyfill https://github.com/majido/high-resolution-timestamp-polyfill/blob/master/translate-timeStamp.js

Hmm, that is interesting. I don't think we need that level of complexity for the current hosts, but will need to keep that in mind.

kerrynf:
> > This looks good, but I just want to note that this wasn't always the case (before 2016), but probably not a case we need to consider?
> > See e.g. this polyfill https://github.com/majido/high-resolution-timestamp-polyfill/blob/master/translate-timeStamp.js
> 
> Hmm, that is interesting. I don't think we need that level of complexity for the current hosts, but will need to keep that in mind.

Need to learn how to use GitHub. Apparently you don't resolve comment threads, but you can abandon the entire PR? :)

PR #65, opened 2023-01-23
Add `teams_people` EntityType to Nova Types

cymushier:
As we are looking at integrating People Experience into Teams, we need the `EntityType` to represent this integration in the commanding.


kerrynf:
Nova Commanding is deprecated in favor of Eventing. Once we have all existing code fully migrated, we can delete this pathway. You shouldn't need to use this pathway for any new integrations.

PR #68, opened 2023-03-14
refactor: use type imports and recommended lint rules, bump dependencies

sjwilczynski:
In scope of this PR we make set of our eslint/tsconfig rules a little stricter:
* we add recommended config from `@typescript-eslint` and fix major issues
* we force using type import where applicable ([importsNotUsedAsValues](https://www.typescriptlang.org/tsconfig#importsNotUsedAsValues)) - that is useful particularly for modern build tools (when [isolatedModules](https://www.typescriptlang.org/tsconfig#isolatedModules) is also set)  and from dev perspective allows to easier understand what is just needed for type and what during runtime
* some more reasonable defaults in tsconfig ([forceConsistentCasingInFileNames](https://www.typescriptlang.org/tsconfig#forceConsistentCasingInFileNames), [noUnusedLocals](https://www.typescriptlang.org/tsconfig#noUnusedLocals), [noUnusedParameters](https://www.typescriptlang.org/tsconfig#noUnusedParameters))
* bump eslint and typescript related dependencies

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -20,5 +20,8 @@
       }
     ]
   },
-  "extends": ["plugin:prettier/recommended"]
+  "extends": [
+    "plugin:@typescript-eslint/recommended",
+    "plugin:prettier/recommended"
+  ]
 }
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -10,9 +10,15 @@
     "node": true
   },
   "ignorePatterns": ["**/coverage", "**/lib", "**/temp"],
-  "plugins": ["prettier"],
+  "plugins": ["@typescript-eslint", "prettier"],
   "rules": {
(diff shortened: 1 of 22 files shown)


freiksenet:
We don't depend on Typescript here, right? (if we do we need to upgrade it). Looks good otherwise.

sjwilczynski:
@freiksenet I bumped TS but needed to bump apollo client and adjust pipeline to use node 16. Let me know if that is ok :)

freiksenet:
if you are bumping versions, change type can't be none - we need major I think.

sjwilczynski:
> if you are bumping versions, change type can't be none - we need major I think.

Good callout, I adjusted the change file for `nova-react-test-utils`, kept the rest as none, as in those packages there were no bumps

PR #71, opened 2023-03-15
Add useRefetchableFragment and usePaginationFragment hooks

dragoshomner:
By bumping [apollo-react-relay-duct-tape](https://github.com/microsoft/graphitation/tree/main/packages/apollo-react-relay-duct-tape), now we are able to use useRefetchableFragment and usePaginationFragment hooks, but they weren't exposed by nova-facade. Therefore, the scope of this PR is to add these hooks, following the same signatures as graphitation.

In this scope, I did the following changes:
- added both hooks to packages/nova-react/src/graphql/hooks.ts, where they used the same signature from graphitation: [useRefetchableFragment](https://github.com/microsoft/graphitation/blob/44957aeecdcfeeae0bb5ceb49ef2fe6293b856dd/packages/apollo-react-relay-duct-tape/src/hooks.ts#:~:text=export%20function-,useRefetchableFragment,-%3C) and [usePaginationFragment](https://github.com/microsoft/graphitation/blob/44957aeecdcfeeae0bb5ceb49ef2fe6293b856dd/packages/apollo-react-relay-duct-tape/src/hooks.ts#:~:text=export%20function-,usePaginationFragment,-%3C)
- for both hooks I call the implementation from the host app and return an error if it's not available, similar to other hooks
- to have a typing consistency, I reuse the same types from graphitation, that were added to packages/nova-react/src/graphql/types.ts. Even if `FetchPolicy` is used for `useLazyLoadQuery`, it's not the scope of this PR to change `options` -> `fetchPolicy` type (it can be done in another one if it's required reusing the same `FetchPolicy` type definition. I chose to add these types in type.ts to keep hooks.ts as clean as possible.
- updated the `nova/types` with the new nook definitions in packages/nova-types/src/nova-graphql.interface.ts
- added unit tests in packages/nova-react/src/graphql/hooks.ts


alloy:
Looks good, just awaiting confirmation that this has been verified to worked in usage.

packages/nova-react/src/graphql/hooks.test.tsx
```diff
@@ -126,6 +132,126 @@ describe(useFragment, () => {
...
+      isLoadingNext: false,
+      isLoadingPrevious: false,
+      refetch: jest.fn(),
+    };
+
+    const graphql: NovaGraphQL = {
+      usePaginationFragment: jest.fn(() => mockedResponse),
+    };
+
+    const fragment = {} as unknown as GraphQLTaggedNode;
```
pasubhas, line 209: Instead of `unknown` as `GraphQlTaggedNode` , use this -> `{ __brand: "GraphQLTaggedNode" }`

kerrynf (approved):

PR #72, opened 2023-03-15
feat(nova-react-test-utils): add storybook decorator for nova

### LINKED ISSUE #17: Storybook support for Nova GraphQL
Apollo has an addon for Storybook: https://storybook.js.org/addons/storybook-addon-apollo-client/

We should provide similar functionality for components built with the Nova facade.

- loading state
- error state
- mock data inputs for various queries

sjwilczynski:
In scope of this PR we add a nova storybook decorator that should cover #17. The whole purpose of the decorator is to expose all the power of existing nova test utils (based on [apollo-mock-client](https://github.com/microsoft/graphitation/tree/main/packages/apollo-mock-client) and [mock-payload-generator](https://github.com/microsoft/graphitation/tree/main/packages/graphql-js-operation-payload-generator)) also inside storybook. For usage details check the README from nova/react-test-utils package.

Some important implementation considerations:
* the main point of the decorator was to allow same level of flexibility as nova/react-test-utils in unit test. So the decorator has `enableQueuedMockResolvers: false` mode, which doesn't generate any mocks by default but instantiates a nova environment for your component which you can get using `getEnvForStory` inside `play` function and perform any actions you'd normally perform during unit tests
* however, most of the time when storybook is run, people don't want to react to each interaction programatically as in unit tests but also have a possibility to interact with a story displayed in Storybook. To achieve that `enableQueuedMockResolvers: true` mode can be used to queue all resolvers up front, so that users can interact with the story live and also makes it simpler in case someone is interested in just resolving a single query/fragment. You can also pass custom resolvers using `resolvers` field which can be strongly typed with `TypeMap`
* we added `examples` folder, so that we can test our decorator on real components. In future the number of examples should increase, including leveraging decorator for fragment only based component, or components leveraging pagination/refetchable  fragments, or subscriptions.
* We thought if storybook deps should be added as peer or normal dependencies in `nova-react-test-utils`. We landed on normal deps similarly how `@apollo/client` to graphitation utilities are added. On clean repo depending on @storybook/addon-actions adds around ~500 lines in lock file, which is acceptable.
* We compared our implementation with https://github.com/iamchanii/storybook-addon-relay/blob/main/src/decorators/withRelay.tsx but instead of adding an option to wrap a story inside a query to be able to leverage MockPayloadGenerator, a preferred solution is to upstream add a possibility to generate mock data for fragment. Let's fix that in graphitation mock payload generator!

TODO: 
* [x] update code to rely on story id instead of name when https://github.com/storybookjs/storybook/pull/22471 is released

Closes #17


packages/examples/.storybook/preview.ts
```diff
@@ -0,0 +1,15 @@
...
+    controls: {
+      matchers: {
+        color: /(background|color)$/i,
+        date: /Date$/,
+      },
+    },
+  },
+};
+
+export default preview;
```
alloy, line 15: beep boop beep

No idea what this all is for 😄

packages/examples/src/Feedback/FeedbackContainer.stories.ts
```diff
@@ -0,0 +1,104 @@
...
+type Story = StoryObj<typeof meta>;
+
+export const Primary: Story = {
+  parameters: {
+    novaEnvironment: {
+      resolvers: {
+        Feedback: () => sampleFeedback,
+      },
+    },
+  } satisfies NovaEnvironmentDecoratorParameters<TypeMap>,
```
alloy, line 30: Would be nice if we can get rid of this in the future.

packages/examples/src/Feedback/FeedbackContainer.test.tsx
```diff
@@ -0,0 +1,41 @@
...
+    const error = await screen.findByText("Something went wrong");
+    expect(error).toBeInTheDocument();
+  });
+
+  it("should show loading state", async () => {
+    render(<Loading />);
+    const loading = await screen.findByText("Loading...");
+    expect(loading).toBeInTheDocument();
+  });
+});
```
alloy, line 41: Passes, but gives an "act" error for me.

Screenshot 2023-05-13 at 01 22 44

packages/nova-react-test-utils/README.md
```diff
@@ -2,4 +2,170 @@
...
+
+Sure, please check [unit tests file](../examples/src/Feedback/FeedbackContainer.test.tsx) to see how stories can be leveraged inside unit tests, using [composeStories](https://github.com/storybookjs/testing-react) to treat storybook as the single source of truth for all the config/setup needed to test your component.
+
+#### How can I use the test utils (both in stories and tests) for components which don't fire graphql operations but rely on graphql fragments only?
+
+If your component, doesn't make any graphql operations and only calls `useFragment` you can still use the `NovaMockEnvironmentProvider` but won't benefit immediately from the apollo-mock-client, as it only intercepts real operations. But that also means that you can pass the fragment value through component props. If you are using Typescript, the data masking of the compiler will prevent you from hacing properly typed fragments. At this point in time, `MockPayloadGenerator` doesn't support yet generating mock data for fragments, but it is planned to be added in the future. For now you can either hardcode fragment value or wrap your component in additional layer which will trigger the graphql query.
```
alloy, line 171:
```suggestion
If your component, doesn't make any graphql operations and only calls `useFragment` you can still use the `NovaMockEnvironmentProvider` but won't benefit immediately from the apollo-mock-client, as it only intercepts real operations. But that also means that you can pass the fragment value through component props. If you are using Typescript, the data masking of the compiler will prevent you from having properly typed fragments. At this point in time, `MockPayloadGenerator` doesn't support yet generating mock data for fragments, but it is planned to be added in the future. For now you can either hardcode fragment value or wrap your component in additional layer which will trigger the graphql query.
```

packages/nova-react-test-utils/README.md
```diff
@@ -2,4 +2,170 @@
...
+const button = await screen.findByRole("button", { name: "Click me" });
+userEvent.click(button);
+expect(environment.eventing.bubble).toHaveBeenCalled();
+```
+
+Let's break it down:
+
+`createMockEnvironment` creates a Nova environment object that can be used with the `NovaMockEnvironmentProvider` and has mocks instantiated for each piece of the facade interface (eventing, commanding and graphql). `createMockEnvironment` takes a schema as an argument. The schema is a GraphQL schema that is later used to generate mock data for the GraphQL mock.
+
+The `environment.graphql.mock` mock provides an API to act on operations issued by a component tree. For details on the API see [apollo-mock-client docs](https://github.com/microsoft/graphitation/tree/main/packages/apollo-mock-client). Together with mock client, which intercepts all operations, for mock data generation one can use `MockPayloadGenerator` whose generate function is leveraging [graphql-js-operation-payload-generator](https://github.com/microsoft/graphitation/tree/main/packages/graphql-js-operation-payload-generator).
```
alloy, line 42: The docs for these need to be created/updated, but in the meantime it _might_ be helpful to link to https://relay.dev/docs/guides/testing-relay-components/

I'm on the fence, though. What do you think?

alloy (approved):
Absolutely amazing! 🤩 

Really solid job, @sjwilczynski—thanks 👏

packages/examples/.storybook/preview.ts
sjwilczynski: No worries [main.ts](https://storybook.js.org/docs/react/configure/overview#configure-your-storybook-project) and [preview.ts](https://storybook.js.org/docs/react/configure/overview#configure-story-rendering) configure your Storybook. The first one is responsible for more low level config (what UI framework, what features, addons, etc.), while in preview one can configure global decorators and parameters. The current `preview.ts` was just autogenerated by Storybook, I will see if it makes sense to make it even simpler :)

packages/examples/src/Feedback/FeedbackContainer.stories.ts
sjwilczynski: of `satisfies`? Then probably no, we won't get rid of it. We still need a way to strongly type `parameters` (by default it's just generic `Record`) so we either need to use satisfies or adjust story type itself

packages/examples/src/Feedback/FeedbackContainer.test.tsx
sjwilczynski: Will check, I was running `yarn ci` locally, that's why I haven't noticed it :)

packages/nova-react-test-utils/README.md
sjwilczynski: Hmm fair point, to at least mention it's built on top of it. I just didn't like relay docs because they rely heavily on `enzyme`.

sjwilczynski: Updated, let me know what you think :)

packages/examples/src/Feedback/FeedbackContainer.test.tsx
sjwilczynski: Fixed!

packages/examples/src/Feedback/FeedbackContainer.stories.ts
sjwilczynski: image

packages/examples/.storybook/preview.ts
alloy: Nah no worries, I was just mostly having a bit of fun 😊

packages/examples/src/Feedback/FeedbackContainer.stories.ts
alloy: Yeah the satisfies part. I figured this was the reason, but I’m thinking that it should be very possible to explain users how they can create their own StoryObj type that composed the normal StoryObj and a resolvers map with their schema.

packages/nova-react-test-utils/README.md
alloy: Yup, looks great 👌

packages/examples/src/Feedback/FeedbackContainer.stories.ts
alloy: (To be clear, this is not a blocker!)

sjwilczynski: One can always do `const Primary: StoryObj & {parameters: NovaEnvironmentDecoratorParameters<TypeMap>}` but I personally like satisfies part :)

packages/nova-react-test-utils/README.md
```diff
@@ -125,7 +124,7 @@ export const LikeFailure: Story = {
 };
 ```
 
-This time not resolvers are queued up front so inside [play](https://storybook.js.org/docs/react/writing-stories/play-function#page-top) one needs to manually resolve/reject graphql operations. To get the environment created for this specific story one can use `getEnvForStory` function. Later similarly to examples for unit test, full customization power of apollo-mock-client is available.
+This time not resolvers are queued up front so inside [play](https://storybook.js.org/docs/react/writing-stories/play-function#page-top) one needs to manually resolve/reject graphql operations. To get the environment created for this specific story one can use `getNovaEnvironmentDecorator` function. Later similarly to examples for unit test, full customization power of apollo-mock-client is available.
```
alloy, line 127: > This time **all** resolvers …

?

alloy: Ah just read in context of the whole document, so it should be:

> This time **no** resolvers

sjwilczynski: Fixed

alloy (approved):

PR #73, opened 2023-03-24
deps(nova-react, nova-react-test-utils): bump React to 17

sjwilczynski:
We are using pretty old version of react in nova-facade. Seems though that our main consumers (TMP, 1JS, App Bar) are all already at version 17. Let's bump.

Please let me know if there are any reasons we shouldn't but at least in 1JS we will be able to remove some resolutions thanks to that :)


alloy:
TMP wasn't on 17 at the time, but now it is 👍

alloy (approved):

PR #74, opened 2023-03-24
deps: bump root dependencies

sjwilczynski:
When working on #72 I noticed some dependencies are pretty old and it caused issues with for example `satisfies` syntax in typescript. In scope of this PR we bump root repo dependencies and apply necessary migrations.


alloy (approved):
Ace, thanks! 🙌

PR #75, opened 2023-03-30
Update package.json files with exports.types and repository entries

MLoughry:
When building with strict mode with Typescript 5.0 using [`resolvePackageJsonExports`](https://devblogs.microsoft.com/typescript/announcing-typescript-5-0/#resolvepackagejsonexports), Typescript will ignore the root `types` field and look for `types` in the `exports` entry being imported. If it's not present, then no type information is imported, and it will throw if `strict` mode is enabled.

The simple fix here is to mirror the root `types` field in `exports["."].types`.

Additionally, I added links to this repository in the `package.json`, as I had to ask around to find the repository.


alloy:
Can you run `yarn change` to generate the release files? These should be patch level version bumps.

PR #79, opened 2023-05-15
Add Nova Scheduling Provider

drewatk:
Adds a new provider, context, and interfaces for a scheduling service to be exposed as a nova context. This will allow react components to schedule units of work to an app's scheduling services.

The interfaces follow TMP's implementation, with generic string ids to track `scheduledById` and `jobId`. The interface allows cancelling the job via the job handle returned from  Priorities follow those defined by the _Prioritized Task Scheduling_ API within chrome: https://wicg.github.io/scheduling-apis/#sec-task-priorities.


packages/nova-types/src/nova-scheduling.interface.ts
```diff
@@ -0,0 +1,25 @@
...
+   * The requested priority of the job.
+   */
+  priority: JobPriority;
+  /**
+   * The identifier of the type of job. Note that this is not globally
+   * unique amongst all instance of a job type.
+   */
+  id: string;
+  /**
+   * The identifier of the type of code which scheduled the job. Not that this is
```
kerrynf, line 19:
```suggestion
   * The identifier of the type of code which scheduled the job. Note that this is
```

packages/nova-types/src/nova-scheduling.interface.ts
```diff
@@ -0,0 +1,25 @@
...
+type IJob<TResult> = (() => Promise<TResult>) | (() => TResult);
+
+export interface IJobMetaData {
+  /**
+   * The requested priority of the job.
+   */
+  priority: JobPriority;
+  /**
+   * The identifier of the type of job. Note that this is not globally
+   * unique amongst all instance of a job type.
```
kerrynf, line 15:
```suggestion
   * unique amongst all instances of a job type.
```

packages/nova-react/src/scheduling/nova-scheduling-provider.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+interface NovaSchedulingProviderProps {
+  scheduling: NovaScheduling;
+}
+
+export const NovaSchedulingProvider: React.FunctionComponent<
+  NovaSchedulingProviderProps
+> = ({ children, scheduling }) => {
+  return (
+    <NovaSchedulingContext.Provider value={scheduling}>
```
kerrynf, line 16: Is it possible to add a default value for scheduling to support hosts which simple needs? We don't have any examples in this PR of how we expect a host to hook this, or a default implementation... :)

packages/nova-types/src/nova-scheduling.interface.ts
```diff
@@ -0,0 +1,25 @@
+export interface NovaScheduling {
```
kerrynf, line 1: Take a look at the comments on the other interface files - let's add some inline documentation around what this is and how to use it. At the moment, this is not clear on how a host would fulfill its duty to a component that is using this API.

Without a default implementation, it will also be difficult for component builders to use the API in their development environments - we will need to think through what the simplest path is for someone who wants to support this in Storybook for example...
/**
 * Describes the eventing contract that is used to communicate between UI components and their logical parent within a host application.
 *
 * Usage: components should create small independent packages that define the events they generate using this contract.
 * Logical parents can take a dependency on the event package, and use it to translate the events into appropriate actions within the larger application.
 */

packages/nova-types/src/nova-scheduling.interface.ts
```diff
@@ -0,0 +1,25 @@
...
+  schedule(job: IJob<any>, meta: IJobMetaData): number;
+  cancel(jobId: number): void;
+}
+
+type IJob<TResult> = (() => Promise<TResult>) | (() => TResult);
+
+export interface IJobMetaData {
```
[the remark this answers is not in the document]
drewatk: The interface should be kept generic to provide scheduling implementation flexibility for consumers. 

Operations that aren't currently supported by the API are changing priority and exposing a completion promise. My thought was to keep the API simply at the minimal set of features needed, and then we can add more as required.

sjwilczynski:
Should we add scheduling provider in [test utils](https://github.com/microsoft/nova-facade/blob/dc359886e3245ef4eee4331409d0eed98e67616b/packages/nova-react-test-utils/src/nova-mock-environment.tsx#LL33C14-L33C41) so that it has fully mocked nova environment?

sjwilczynski:
Would be also awesome to add some example component relying scheduling provider to `packages/examples` :)

PR #85, opened 2023-10-06
Add unmountEventing through the NovaEventingProvider

drewatk:
We've discovered a need to have two versions of eventing at once during component unmounts. Useeffect cleanup functions need to use the old version of eventing to get the old context to ensure proper routing of events. This change will allow hosts to maintain two eventing references, with the unmount eventing switching out later to keep the reference to the old execution context before unmount.

This change adds a useNovaUnmountEventing hook in the nova eventing provider. It is supplied by a new `unmountEventing` property. If the new property is not supplied, the new hook will use the existing eventing reference, to maintain backwards compatibility with nova hosts which do not need this functionality.

### DIFF
--- a/packages/nova-react/src/eventing/nova-eventing-provider.test.tsx
+++ b/packages/nova-react/src/eventing/nova-eventing-provider.test.tsx
@@ -19,6 +19,8 @@ import { InputType } from "@nova/types";
 
 import * as ReactEventSourceMapper from "./react-event-source-mapper";
 
+const { useEffect } = React;
+
 jest.mock("./react-event-source-mapper");
 
 describe("useNovaEventing", () => {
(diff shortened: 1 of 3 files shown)


packages/nova-react/src/eventing/nova-eventing-provider.tsx
```diff
@@ -4,10 +4,16 @@ import { InputType } from "@nova/types";
...
+const NovaEventingContext = React.createContext<INovaEventingContext>({});
+
+interface INovaEventingContext {
+  eventing?: NovaReactEventing;
+  unmountEventing?: NovaReactEventing;
+}
 
 interface NovaEventingProviderProps {
   eventing: NovaEventing;
+  unmountEventing?: NovaEventing; // Optional version of eventing to use when unmounting, defaults to eventing if not supplied
```
kerrynf, line 17: Can we change the comment to a doc comment over the prop?

drewatk: Updated

packages/nova-react/src/eventing/nova-eventing-provider.test.tsx
```diff
@@ -260,3 +278,63 @@ describe("NovaReactEventing exposes 'generateEvent'", () => {
...
+    render(
+      <NovaEventingProvider
+        eventing={eventing}
+        unmountEventing={unmountEventingMock}
+        reactEventMapper={mapper}
+      >
+        <TestPassedContextComponent />
+      </NovaEventingProvider>,
+    );
+  });
```
kerrynf, line 339: Can we add a case for the scenario we are adding this for? (Unmounting?)

packages/nova-react/src/eventing/nova-eventing-provider.tsx
```diff
@@ -47,57 +57,92 @@ export interface NovaReactEventing {
...
 
-    return (
-      <NovaEventingContext.Provider value={reactEventing}>
-        {children}
-      </NovaEventingContext.Provider>
-    );
-  };
+  return (
+    <NovaEventingContext.Provider
+      value={{ eventing: reactEventing, unmountEventing: reactUnmountEventing }}
```
blakecscott, line 94: As long as the provider does not re-render, this will not be an issue. However, every time the NovaEventingContext.Provider re-renders, all children will re-render, since the object reference assigned to the value prop is created inline. Please consider stabilizing this object reference with a memo.

blakecscott (approved):

blakecscott (approved):

kerrynf (approved):

PR #88, opened 2023-12-05
Add  support for Relay 13+ generated code by adding $fragmentSpreads support

Markionium:
Relay 16 generates the following fragment type

```ts
export type JobTitleFragment$key = {
  readonly " $data"?: JobTitleFragment$data;
  readonly " $fragmentSpreads": FragmentRefs<"JobTitleFragment">;
};
```

Our hooks currently only support the older `" $fragmentRefs"` format that looks as follows.

```ts
export type JobTitleFragment$key = {
  readonly " $data"?: JobTitleFragment$data;
  readonly " $fragmentRefs": FragmentRefs<"JobTitleFragment">;
};
```

This means that when generating code with newer versions of the relay-compiler we can not use those generated fragments with `useFragment` and will get the following error.

```
Argument of type 'JobTitleFragment$key' is not assignable to parameter of type 'Readonly<{ " $data"?: unknown; " $fragmentRefs": unknown; }>'.
  Property '" $fragmentRefs"' is missing in type 'JobTitleFragment$key' but required in type 'Readonly<{ " $data"?: unknown; " $fragmentRefs": unknown; }>'.ts(2345)
types.d.ts(32, 5): '" $fragmentRefs"' is declared here.
```

This PR adds support for both variants which makes `useFragment` compatible with with Relay types < 13 and Relay > 13.

### DIFF
--- a/change/@nova-react-8a6bad44-a7eb-4078-8e64-97d0eaeedc62.json
+++ b/change/@nova-react-8a6bad44-a7eb-4078-8e64-97d0eaeedc62.json
@@ -1,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Add support for $fragmentSpreads format",
+  "packageName": "@nova/react",
+  "email": "mark@thedutchies.com",
+  "dependentChangeType": "patch"
+}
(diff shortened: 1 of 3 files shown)


alloy (approved):
Great, thanks!

PR #89, opened 2023-12-06
Allow React 18 to a peerDependency and fix Provider Prop types

Markionium:
When trying to use `@nova/react` with React 18 the typings for the providers are not correct as `children` now has to be explicitly defined on the Props type.

This adds the `children` prop to all the providers but keeps us on React 17 for now as that is what most of the internal projects still use. The fix should however allow React 18 users to use `@nova/react` without any errors.


kerrynf (approved):
Looks like a good thing to do!

PR #91, opened 2024-01-14
Move compiler to separate package

Markionium:
Currently the `@nova/react` package comes with a compiler that helps create the types for the components.  This compiler is based on the older relay compiler. 

We have in some cases migrated to using the new rust based relay compiler. This means the bundled compiler is now no longer used for those projects. 

This PR moves the compiler currently available in @nova/react to a separate package which when needed can be installed separately.

The script name is the same so migration from the bundled compiler to the separate one should be easy.


alloy (approved):
I think we should add a README to the compiler package that mentions this is just there for some backwards compatibility, but that it should be considered deprecated and people should look at duct-tape-compiler instead.

PR #92, opened 2024-02-21
Added optional source to GeneratedEventWrapper

kennetvu:
We are currently integrating Jana into Outlook Web(OWA) and we need a way to catch a NovaEvent and then bubble it to parent provider if we dont want to deal with it. We are doing this by calling `generateEvent(event)`, but with current `generateEvent(event)`, source is generated on each call, which means we will lose metadata about `source` if it already exists in the event. 

By adding a optional field `source`, we can generate only if `source` is not defined.

### DIFF
--- a/packages/nova-react/src/eventing/nova-eventing-provider.tsx
+++ b/packages/nova-react/src/eventing/nova-eventing-provider.tsx
@@ -1,5 +1,10 @@
 import * as React from "react";
-import type { NovaEvent, NovaEventing, EventWrapper } from "@nova/types";
+import type {
+  NovaEvent,
+  NovaEventing,
+  EventWrapper,
+  Source,
+} from "@nova/types";
 import { InputType } from "@nova/types";
 import invariant from "invariant";
 
@@ -37,6 +42,10 @@ export interface ReactEventWrapper {
 
 export interface GeneratedEventWrapper {
   event: NovaEvent<unknown>;
+  /**
+   * Optional details about the originating event like input method and timestamp
+   */
+  source?: Source;
   /**
    * Optional timestamp in milliseconds since epoch format,
    * by default will use Date.now() if override not supplied.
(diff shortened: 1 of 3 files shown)


sjwilczynski:
LGTM to me but let's add a test exactly for our use case

sjwilczynski (approved):

packages/nova-react/src/eventing/nova-eventing-provider.tsx
```diff
@@ -142,7 +151,7 @@ const generateEventing =
     generateEvent: (eventWrapper: GeneratedEventWrapper) => {
       const mappedEvent = {
         event: eventWrapper.event,
-        source: {
+        source: eventWrapper.source ?? {
```
kerrynf, line 154: This change is a little weird, because we already support the timeStampOverride here, which is now completely ignored if the optional source is passed in.

Markionium: This kind of comes down to what we want the behaviour to be? :D 

In this case it seems we just want to "forward" the event to a handler further up in the tree. In that case should the overrides apply as initially the event was likely created using `bubble`?

But perhaps we need another way to just "bubble" those events further? It's almost like just doing this?
```ts 
forward: (eventWrapper: EventWrapper) => {
  return eventingRef.current.bubble(eventWrapper);
},
```

kennetvu: > This change is a little weird, because we already support the timeStampOverride here, which is now completely ignored if the optional source is passed in.

We can also turn it around to check for timeStampOverride first, and then source. otherwise fallback to default which is `{inputType: InputType.programmatic, timeStamp:  Date.now()}`

kennetvu: > This kind of comes down to what we want the behaviour to be? :D
> 
> In this case it seems we just want to "forward" the event to a handler further up in the tree. In that case should the overrides apply as initially the event was likely created using `bubble`?
> 
> But perhaps we need another way to just "bubble" those events further? It's almost like just doing this?
> 
> ```ts
> forward: (eventWrapper: EventWrapper) => {
>   return eventingRef.current.bubble(eventWrapper);
> },
> ```

this is kind of what we wanted to achieve.

sjwilczynski:
@kennetvu let's close this in favor of #93

sjwilczynski:
Closing as intercepter pattern was implemented in #93

PR #93, opened 2024-03-05
Nova Event interceptor

kerrynf:
Add the ability for a component lower in the tree to intercept events and possibly process them directly, while maintaining the processing overrides set by the top level provider.

### DIFF
--- a/packages/nova-react/src/eventing/nova-eventing-provider.test.tsx
+++ b/packages/nova-react/src/eventing/nova-eventing-provider.test.tsx
@@ -28,6 +28,7 @@ describe("useNovaEventing", () => {
   let eventing: NovaEventing;
   let prevWrappedEventing: NovaReactEventing;
   let eventCallback: () => void;
+  const renderSpy = jest.fn();
   const initialChildren = "initial children";
   const updatedChildren = "updated children";
 
@@ -43,6 +44,7 @@ describe("useNovaEventing", () => {
     prevWrappedEventing = undefined as unknown as NovaReactEventing;
 
     TestComponent = ({ childrenText }) => {
+      renderSpy();
       const wrappedEventing: NovaReactEventing = useNovaEventing();
       expect(wrappedEventing).toBeDefined();
       expect(wrappedEventing).not.toBe(eventing);
@@ -110,6 +112,8 @@ describe("useNovaEventing", () => {
       initialChildren,
     );
 
+expect(renderSpy).toHaveBeenCalledTimes(1);
+
     wrapper.rerender(
       <NovaEventingProvider
         children={<TestComponent childrenText={updatedChildren} />}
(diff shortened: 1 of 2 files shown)


packages/nova-react/src/eventing/nova-eventing-provider.tsx
```diff
@@ -133,20 +192,37 @@ const generateEventing =
...
-    bubble: (eventWrapper: ReactEventWrapper) => {
+    bubble: async (eventWrapper: ReactEventWrapper) => {
       const mappedEvent: EventWrapper = mapperRef.current(eventWrapper);
-      return eventingRef.current.bubble(mappedEvent);
+      if (!interceptorRef) {
+        return eventingRef.current.bubble(mappedEvent);
+      }
+
+      let eventToBubble: EventWrapper | undefined = mappedEvent;
+      eventToBubble = await interceptorRef?.current(mappedEvent);
```
sjwilczynski, line 207: interceptorRef is not null here, right? We don't need `?.`

sjwilczynski, line 207: When can `eventToBubble` be null/undefined here?

sjwilczynski: Should we return undefined from interceptor handler to make sure the processed event is not bubbled up? And otherwise just return event it got, right?

packages/nova-react/src/eventing/nova-eventing-provider.test.tsx
```diff
@@ -43,6 +44,7 @@ describe("useNovaEventing", () => {
     prevWrappedEventing = undefined as unknown as NovaReactEventing;
```
sjwilczynski, line 44: Let's add a test which is using interceptor and making sure one event is processed and the other moved up

packages/nova-react/src/eventing/nova-eventing-provider.tsx
```diff
@@ -133,20 +192,37 @@ const generateEventing =
...
           timeStamp: eventWrapper.timeStampOverride ?? Date.now(),
         },
       };
-      return eventingRef.current.bubble(mappedEvent);
+      if (!interceptorRef) {
+        return eventingRef.current.bubble(mappedEvent);
+      }
+
+      let eventToBubble: EventWrapper | undefined = mappedEvent;
+      eventToBubble = await interceptorRef?.current(mappedEvent);
```
pasubhas, line 224: Formatting is a little off in the whole page but seems like a reasonable addition.

Could you please add an example for using it in README?

packages/nova-react/src/eventing/nova-eventing-provider.tsx
kennetvu: I think its complaining because line 202 does not check for `.current`

sjwilczynski:
I addressed my comments and added tests/Readme in #96, would be awesome @kerrynf if you decided on what you want to take from that branch, use it and get PR in :)

PR #94, opened 2024-03-06
Allow fragment refs to be null or undefined

Markionium:
Relay allows for fragment references  to be `null` or `undefined`. Our typings have been stricter where they do not allow `null` or `undefined`. 

This causes us to miss out on things like `@skip` and `@include`.

This PR adds the ability to pass `null` or `undefined` as the fragment reference to `useFragment` `usePaginationFragment` or `useRefetchableFragment`.

### DIFF
--- a/packages/nova-react/src/graphql/hooks.test.tsx
+++ b/packages/nova-react/src/graphql/hooks.test.tsx
@@ -4,6 +4,7 @@
 
 import React from "react";
 import { render, screen } from "@testing-library/react";
+import { renderHook } from "@testing-library/react-hooks";
 
 import { NovaGraphQLProvider } from "./nova-graphql-provider";
 import type { NovaGraphQL } from "@nova/types";
@@ -17,6 +18,9 @@ import {
 } from "./hooks";
 import type { GraphQLTaggedNode } from "./taggedNode";
 import type { FragmentRefs } from "./types";
+
+type NotNull<T> = null extends T ? false : true;
+type NotUndefined<T> = undefined extends T ? false : true;
 
 describe(useLazyLoadQuery, () => {
   it("ensures an implementation is supplied", () => {
@@ -67,7 +71,7 @@ describe(useLazyLoadQuery, () => {
     expect(graphql.useLazyLoadQuery).toHaveBeenCalledWith(
       query,
       {},
-      { context: { callerInfo: "subject-with-query" } }
+      { context: { callerInfo: "subject-with-query" } },
     );
     expect(screen.getByText("some-data")).toBeDefined();
   });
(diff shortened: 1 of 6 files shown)


packages/nova-react/src/graphql/hooks.ts
```diff
@@ -193,6 +252,39 @@ export function usePaginationFragment<
   isLoadingNext: boolean;
   isLoadingPrevious: boolean;
   refetch: RefetchFn<TQuery["variables"]>;
+};
+export function usePaginationFragment<
```
sjwilczynski, line 256: How is this definition different from one below?

Markionium: If you mean the one with the implementation then it isn't different. But afaik you can't have the implementation serve as one of the definitions.

sjwilczynski: Aha, one is definition the other implementation

sjwilczynski (approved):

packages/nova-react/src/graphql/hooks.test.tsx
```diff
@@ -152,9 +156,87 @@ describe(useFragment, () => {
...
+    type SomeFragment$data = { someKey: string };
+    type SomeFragment$key = {
+      readonly " $data"?: SomeFragment$data;
+      readonly " $fragmentRefs": FragmentRefs<"SomeFragment">;
+    };
+
+    const fragment = {} as unknown as GraphQLTaggedNode;
+    const opaqueFragmentRef = {} as unknown as SomeFragment$key | null;
+
+    () => {
```
alloy, line 192: You can start a new scope with just the curly braces, which might make it slightly less confusing—I spend a few sec trying to understand what you were trying to do with the function.

No biggie, though.

packages/nova-react/src/graphql/hooks.test.tsx
```diff
@@ -152,9 +156,87 @@ describe(useFragment, () => {
...
+    });
+
+    expect(result.current).toBeNull();
+
+    rerender({ fragmentRef: undefined });
+
+    expect(result.current).toBeUndefined();
+  });
+
+  it("return type can be `null`, `undefined` or data", () => {
```
alloy, line 182: Nit: "it return type …" isn't English

packages/nova-react/src/graphql/hooks.test.tsx
```diff
@@ -152,9 +156,87 @@ describe(useFragment, () => {
...
+      const _: ExpectedReturnType = null;
+      const __: ExpectedReturnType = undefined;
+      const ___: ExpectedReturnType = { someKey: "some-data" };
+
+      // Workaround for TS complaining about unused variables
+      void _, __, ___;
+    };
+  });
+
+  it("return type does not include `null` or `undefined`", () => {
```
alloy, line 206: "… when a fragment ref is passed"

Right?

packages/nova-react/src/graphql/hooks.ts
```diff
@@ -153,13 +166,42 @@ export function useFragment<TKey extends KeyType>(
  *                    fragment.
  * @returns The data corresponding to the field selections and a function to perform the refetch.
  */
+export function useRefetchableFragment<
+  TQuery extends OperationType,
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  _TKey extends KeyType,
```
alloy, line 172: This one could do with a comment perhaps, but not necessary if it's captured as a test that would fail without this type parameter definition missing.

packages/nova-react/src/graphql/hooks.ts
```diff
@@ -134,10 +134,23 @@ export function useLazyLoadQuery<TQuery extends OperationType>(
  *                    fragment.
  * @returns The data corresponding to the field selections.
  */
+export function useFragment(
+  fragmentInput: GraphQLTaggedNode,
+  fragmentRef: null | undefined, // This overload makes sure we do not get `unknown` as the return type.
```
alloy, line 139: This comment shouldn't be necessary, as long as there is a test that would fail without this parameter missing?

alloy (approved):
Looks good, just some minor feedback :+1:

packages/nova-react/src/graphql/hooks.test.tsx
Markionium: True, but technically in the jest report it turns into `useFragment return type ...` ;)

Markionium: Will fix though as it doesn't read a nice here :)

packages/nova-react/src/graphql/hooks.test.tsx
Markionium: I used the function to not actually call `useFragment` which will give the invalid hook error ;) So seems that wasn't actually clear after all, I'll fix it.

packages/nova-react/src/graphql/hooks.ts
Markionium: Added a test :)

PR #96, opened 2024-03-07
Continuation of #93

sjwilczynski:


packages/nova-react/src/eventing/nova-eventing-provider.test.tsx
```diff
@@ -380,3 +390,138 @@ describe("useUnmountEventing", () => {
...
+    );
+
+    expect(() => render(<InterceptorTestComponent />)).toThrow(
+      "Nova Eventing provider must be initialized prior to creating NovaEventingInterceptor!",
+    );
+  });
+
+  it("intercepts the event and bubbles up the event when interceptor returns the event anyway", async () => {
+    const interceptor = (eventWrapper: EventWrapper) => {
+      if (eventWrapper.event.originator === "toBeIntercepted") {
```
kerrynf, line 507: Can we add the expect.assertion checks here? I don't want something to break and mean they pass by timing out...

sjwilczynski: Unfortunately, `waitFor` and `expect.asseritons` are not compatible. `waitFor` calls callback in timeout until it succeeds and because it has expect inside, the number of calls to expect is unstable inside test, depending on how fast async operations complete. We need to leave it as is and have `expect.assertions` only in tests where we don't call `waitFor`

kerrynf (approved):

PR #97, opened 2024-03-26
Use FetchPolicy in useLazyLoadQuery

iukondra:


packages/nova-react/src/graphql/types.ts
```diff
@@ -82,11 +82,17 @@ export type PaginationFn = (
...
 ) => Disposable;
 
-type FetchPolicy =
+export type FetchPolicy =
   | "store-or-network"
   | "store-and-network"
   | "network-only"
-  | "store-only";
+  | "store-only"
+  | "cache-first"
```
sjwilczynski, line 90: What are the new fetch policies? Are there from apollo-client? I think we should just use Relay ones as [duct tape does translation](https://github.com/microsoft/graphitation/blob/main/packages/apollo-react-relay-duct-tape/src/convertFetchPolicy.ts). Right, @alloy ?

iukondra: Apollo fetch policies. If we don't need to support them, I'm happy to remove them

alloy: Yes, only the Relay ones, and specifically the ones captured in the duct-tape transform 👍

dragoshomner (approved):
Looks good to me. Please wait Eloy's review before completing it.

sjwilczynski (approved):

alloy (approved):

PR #98, opened 2024-05-08
feat: bump storybook to latest 7.x and adjust improts

sjwilczynski:
In newer version of storybook importing `makeDecorator` from `@storybook/addons` should be replaced with import from `@storybook/preview-api` (which is also compatible with Storybook 8)

In scope of this PR we:
* adjust import for `makeDecorator` and fix types that got renamed
* bump storybook packages to 7.6.19
* start using @storybook/test instead of @storybook/testing-library
* bump lage to latest
* we set `skipLibCheck: true` in tsconfig to avoid errors from `node_modules`


kerrynf:
Looks good to me!

kerrynf (approved):

PR #100, opened 2024-05-17
Fix RefetchOptions onComplete function type

mwangi-njuguna:
`RefetchOptions` type has an incorrect signature and causes linting issues on `onComplete` function usage.


kerrynf:
Looks good to me.

kerrynf (approved):

PR #101, opened 2024-07-23
feat: bump react to version 18 and run all tests/stories in strict mode

sjwilczynski:
As I think all consumers of recent versions of nova are already on react 18, in this PR we make sure we have same version in the repo. In scope of this PR:
* we bump internal dependencies to react 18 and aligned peer deps on allowing it
* we bump testing-library packages which was also needed with react bump
* we enable strictMode for tests and stories
* we bump node version for PR job to 18.x


Markionium (approved):

packages/nova-react/src/eventing/nova-eventing-provider.test.tsx
```diff
@@ -113,7 +105,8 @@ describe("useNovaEventing", () => {
       initialChildren,
     );
 
-    expect(renderSpy).toHaveBeenCalledTimes(1);
+    // called twice on each render due to strict mode
```
kerrynf, line 108: We should probably make this clearer - strict mode is explicitly rerendering the component to help expose bugs. This won't happen in production. (The goal of these tests is to make sure we have a stable context... :)) (https://react.dev/learn/synchronizing-with-effects#how-to-handle-the-effect-firing-twice-in-development)

kerrynf (approved):

packages/nova-react/src/eventing/nova-eventing-provider.test.tsx
sjwilczynski: Ok I updated the tests by encapsulating a bit the logic for render count. I think it makes sense now and tests name are also very descriptive. I will merge the PR as is but @kerrynf please let me know if that is not what you expected and if so I will make a new PR after I come back from vacation on 5th of August

PR #103, opened 2024-08-22
breaking(nova-react-test-utils): add support for Relay

sjwilczynski:
Joint work with @jakobkhansen

In scope of this PR we add support for using Relay client with the Nova decorator and add a feature of specifying `query` needed to be run and way to pass result of that query to the component rendered in Story. Check the documentation that we added in README for more details about the features.

Additionally we changed the structure of the package so that now users have access to same functions for both client but need to import from different paths. For example:
* to import Relay based utils one needs to do `import { getNovaDecorator } from "@nova/react-test-utils"` or `import { getNovaDecorator } from "@nova/react-test-utils/relay"`
*  to import Apollo based utils one needs to do `import { getNovaDecorator } from "@nova/react-test-utils/apollo"`
Same patterns applies for utilities for unit tests

To verify our work, we added more stories and tests in `packages/examples` now showcasing both GraphQL client variants


packages/nova-react-test-utils/package.json
```diff
@@ -11,35 +11,49 @@
     "just": "monorepo-scripts"
   },
   "peerDependencies": {
```
Markionium, line 13: Should we use `https://docs.npmjs.com/cli/v8/configuring-npm/package-json#peerdependenciesmeta` to identify some of these are "optional" depending on the variant you use?

packages/nova-react-test-utils/package.json
```diff
@@ -11,35 +11,49 @@
...
+    "@storybook/addon-actions": "^7.6",
+    "@storybook/preview-api": "^7.6",
     "@storybook/react": "^7.6",
-    "react": "^17.0.2 || ^18"
+    "react": "^17.0.2 || ^18",
+    "@apollo/client": "^3.4",
+    "@graphitation/apollo-mock-client": "^0.11.0",
+    "relay-compiler": "^12.0.0",
```
Markionium, line 20: This should probably be aligned with 16 || 17?

Markionium: Maybe that's not so easy because of duct-tape that needs 12?

So perhaps it's 12 || 16 || 17?

sjwilczynski: I removed it in the end, it was added as peer dep for one of graphitation packages but with them being now in peerDeps of test utils, it is not needed to add them again here

sjwilczynski:
@ansemb 

> I like having @nova/react-test-utils/apollo and @nova/react-test-utils/relay.
> I wouldn't mind having the index.ts only for relay, hence have @nova/react-test-utils (relay) and @nova/react-test-utils/apollo as I think relay should be considered default going forward, though that might also feel weird.

This is how it works right now actually. Because we set the main field as `./lib/relay/index.ts` when importing without subpath it will use relay version. I decided to keep both subpaths for consistency tough

packages/nova-react-test-utils/package.json
```diff
@@ -50,14 +102,14 @@
...
+    "module": "./lib/relay/index.mjs",
     "exports": {
       ".": {
-        "types": "./lib/index.d.ts",
-        "import": "./lib/index.mjs",
-        "require": "./lib/index.js"
+        "import": "./lib/relay/index.mjs",
+        "require": "./lib/relay/index.js",
+        "types": "./lib/relay/index.d.ts"
       }
```
Markionium, line 113: I think you need both `relay` and `apollo` here to make them available for import? The way I understand "exports" to work is that it won't allow other paths to be accessed other than what is here.

```suggestion
       "./relay": {
         "import": "./lib/relay/index.mjs",
         "require": "./lib/relay/index.js",
         "types": "./lib/relay/index.d.ts"
       },
       "./apollo": {
         "import": "./lib/apollo/index.mjs",
         "require": "./lib/apollo/index.js",
         "types": "./lib/apollo/index.d.ts"
       }
      }
```

Markionium: You might not have run into it because it's in `"publishConfig"`? But I'm not sure about that.

packages/nova-react-test-utils/src/relay/index.ts
```diff
@@ -0,0 +1,20 @@
...
+  createNovaRelayMockEnvironment as createMockEnvironment,
+} from "./test-utils";
+export {
+  getRelayOperationName as getOperationName,
+  getRelayOperationType as getOperationType,
+} from "./operation-utils";
+export {
+  getNovaRelayDecorator as getNovaDecorator,
+  getNovaRelayEnvironmentForStory as getNovaEnvironmentForStory,
+} from "./storybook-nova-decorator-relay";
```
Markionium, line 16: Any specific reason for the alias instead of just doing this? I'm assuming they'll never be used together anyways, so this is probably fine? I don't have a strong preference, if you think this is better because of clarity thats fine with me too.

```suggestion
export {
  getOperationName,
  getOperationType,
} from "./operation-utils";
export {
  getNovaDecorator,
  getNovaEnvironmentForStory,
} from "./storybook-nova-decorator-relay";
```

Markionium (commented):
Forgot to complete this yesterday :)

packages/examples/src/relay/Feedback/Feedback.stories.ts
```diff
@@ -0,0 +1,106 @@
...
+          }
+        }
+      `,
+      variables: { id: "42" },
+      getReferenceEntry: (data) => ["feedback", data?.feedback],
+    },
+  } satisfies WithNovaEnvironment<FeedbackStoryQuery, TypeMap>,
+};
+
+export default meta;
```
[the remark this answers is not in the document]
sjwilczynski: It is better, I can change

packages/nova-react-test-utils/src/relay/index.ts
sjwilczynski: In my mind this simplifies searching and replacing as you are able to search for unique text at least within the package. I'd rather keep it that way

packages/nova-react-test-utils/relay.mjs
```diff
@@ -0,0 +1 @@
+export * from "./lib/relay";
```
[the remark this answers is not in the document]
sjwilczynski: Seems so, unless you have an idea how to mitigate issues described in https://github.com/microsoft/nova-facade/pull/103#discussion_r1734860165

packages/nova-react-test-utils/package.json
sjwilczynski: I basically used approach from https://github.com/testing-library/react-testing-library/blob/main/pure.js, just added .js and .mjs variants of these files

packages/nova-react-test-utils/src/relay/storybook-nova-decorator-relay.tsx
```diff
@@ -0,0 +1,85 @@
...
+import { novaGraphql } from "./nova-relay-graphql";
+import { RelayMockPayloadGenerator } from "./test-utils";
+
+import * as React from "react";
+import { RelayEnvironmentProvider } from "react-relay";
+import { createMockEnvironment } from "relay-test-utils";
+
+import type { GraphQLSchema } from "graphql";
+import type { ReactRenderer } from "@storybook/react";
+import type { PlayFunctionContext } from "@storybook/types";
```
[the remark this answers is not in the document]
jakobkhansen: Unrelated, but we should setup auto-organizing imports in this repo like in 1JS 😄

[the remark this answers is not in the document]
sjwilczynski: No, will adjust

packages/nova-graphql-compiler/src/cli.test.js
```diff
@@ -36,5 +36,5 @@ describe("cli", () => {
     );
 
     expect(generatedFiles).toMatchSnapshot();
-  });
+  }, 50000);
```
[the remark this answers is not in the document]
Markionium: It's fine :) I kind of consider this as deprecated anyways.

packages/nova-react-test-utils/package.json
Markionium: > And BTW I verified that using exports field forces us to change moduleResolution to Bundler 

I'm a bit confused about this one, that shouldn't be needed I think.  But I'm totally fine that we keep it this way.

packages/examples/src/relay/Feedback/Feedback.stories.ts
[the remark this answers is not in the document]
sjwilczynski: Done

packages/nova-react-test-utils/src/relay/storybook-nova-decorator-relay.tsx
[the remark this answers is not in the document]
sjwilczynski: Done

packages/nova-react-test-utils/apollo.mjs
```diff
@@ -0,0 +1 @@
+export * from "./lib/apollo";
```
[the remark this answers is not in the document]
sjwilczynski: Closing in favor of the other comment

packages/nova-react-test-utils/relay.mjs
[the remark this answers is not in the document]
sjwilczynski: Closing to keep only one comment on the same thing open

packages/nova-react-test-utils/package.json
sjwilczynski: Resolving then

Markionium (approved):

kerrynf (approved):

PR #109, opened 2024-09-16
fix(react-test-utils): fix remaining integration issues

sjwilczynski:
In #108 we partially fixed some issues, as imports still not work correctly due to wrong imports map and there was an error thrown when opening storybook on jest being used directly. Here we aim to fix these by:
* upgrading typescript to 5.x to be able to use `moduleResolution: "Bundler'` needed for exports map (apart from compiler which needs to stay on 4.x) - this also required bumping @graphitation dependencies and making sure we can use per package tsconfig for tests
* correctly use exports map as it's used by various tools but keep the additional .mjs/.d.ts/.js files to not force consumers to use moduleResolution bundler
* replace @jest/globals with @jest/types to avoid runtime issues in storybook

Verified changes using local storybook and by making a patch in 1JS


packages/nova-graphql-compiler/README.md
```diff
@@ -19,3 +19,5 @@ Then inside your package.json add the compiler to your `scripts`:
   }
 }
 ```
+
```
kerrynf, line 22: This sounds like something we should fix?

packages/examples/tsconfig.jest.json
```diff
@@ -0,0 +1,6 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "noUnusedLocals": false,
```
kerrynf, line 4: // prevents error about graphql import not being used when artifacts are loaded
 
 This looks like it moved from further down, where it had this comment? Is this the reason this is being added here? Should we include the comment?

kerrynf (approved):

packages/nova-graphql-compiler/README.md
sjwilczynski: I agree but who has funding to fix it? I created an issue a while back in https://github.com/microsoft/graphitation/issues/331

packages/examples/tsconfig.jest.json
sjwilczynski: I realized it makes more sense to have local test specific tsconfigs, instead of using a big one. I will make a small PR to improve it even further and add some explanations

PR #110, opened 2024-09-18
fix(examples): improve handling of local jest config and remove RTL clean hack

sjwilczynski:
Following up on a [comment](https://github.com/microsoft/nova-facade/pull/109#discussion_r1763744663) from #109 we improved handling of having package specific configurations:
* we change just scripts to prioritize using package local Jest config if it exists
* we move logic that handles transforms for graphql file and relay compiler generated artifacts to jest config in `examples` package
* we remove a hack we added in `jest.setupAfterEnv` file by using `/pure` imports from RTL

### DIFF
--- a/packages/examples/jest.config.ts
+++ b/packages/examples/jest.config.ts
@@ -1,0 +1,21 @@
+import config from "../../scripts/config/jest.config";
+
+export default {
+  ...config,
+  transform: {
+    "\\.(gql|graphql)$": "@graphql-tools/jest-transform",
+    // Needed for packages/examples to load relay compiler generated artifacts.
+    // It would be better if it could be config local to examples package but haven't found a way to configure it with just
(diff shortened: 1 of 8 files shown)


packages/examples/jest.config.ts
```diff
@@ -0,0 +1,21 @@
+import config from "../../scripts/config/jest.config";
+
+export default {
+  ...config,
+  transform: {
+    "\\.(gql|graphql)$": "@graphql-tools/jest-transform",
+    // Needed for packages/examples to load relay compiler generated artifacts.
+    // It would be better if it could be config local to examples package but haven't found a way to configure it with just
```
kerrynf, line 8: It looks like it is local to the examples package? Am I missing something or does the comment need to be updated?

sjwilczynski: Right, I mindlessly copy pasted, will remove on next PR

PR #113, opened 2024-09-19
fix(nova-react-test-utils): adjust types to properly support pure Relay components

sjwilczynski:
The decorator is written in such a way that in can support pure Relay components, not only Nova. As it adds some features on top of [Relay addon](https://github.com/iamchanii/storybook-addon-relay), making sure using it for pure Relay components can increase consistency and DX of working on Relay/Nova components. In scope of this PR we adjust decorator typing to support using without directly depending on `@nova/react`.


packages/examples/package.json
```diff
@@ -23,7 +23,8 @@
     "@nova/types": "1.5.1",
     "graphql": "^15.5.0",
     "react": "^18.3.1",
-    "react-dom": "^18.3.1"
+    "react-dom": "^18.3.1",
+    "react-relay": "^17.0.0"
```
Markionium, line 27: This is probably better as a peer dependency?

I hope to upgrade to relay 18 soon, which would mean this won't "align" anymore.

packages/nova-react-test-utils/src/shared/storybook-nova-decorator-shared.tsx
```diff
@@ -82,7 +83,11 @@ export function getRenderer(
 ): React.FC<React.PropsWithChildren<unknown>> {
   if (query) {
     const Renderer: React.FC<unknown> = () => {
-      const { data } = useLazyLoadQuery(query, variables);
+      const { data } = useLazyLoadQuery(
+        // There are no consequences of the cast, we do it only to make sure pure relay components can also leverage the decorator
```
Markionium, line 87: During runtime this would be replaced with the runtime document, so it's not actually even this type.

Which when using Nova and Relay should be the same document. But not sure if that is worth adding to the comment?

packages/examples/package.json
Markionium: Oh this is the examples package nvm :D

Markionium (approved):

PR #114, opened 2024-09-20
fix(nova-react-test-utils): improve types for WithoutFragmentRefs

sjwilczynski:
We added a utility `WithoutFragmentRefs` that makes sure Typescript doesn't want us to provide args that are defined on `referenceEntries`. However, the way we wrote the type impacts the `render` function of a Story as well, needing us to either disabled type check there or provide not correct props. In this PR we adjust the type to be `StoryObjWithoutFragmentRefs` which removes the props defined in `referenceEntries` from `args` only, instead of all types in `StoryObj`


packages/nova-react-test-utils/src/shared/types.d-test.ts
Markionium, line 1: This is pretty nice :D

Markionium (approved):

PR #117, opened 2024-09-24
[DO NOT REVIEW] Align useMutation with Relay

DarynaAkhmedova:


packages/nova-react/package.json
```diff
@@ -12,6 +12,7 @@
   },
   "peerDependencies": {
     "react": "^17.0.2 || ^18",
+    "relay-runtime": "12.0.0",
```
Markionium, line 15: We shouldn't need this ? :)

packages/nova-react/src/graphql/hooks.ts
```diff
@@ -10,6 +10,7 @@ import type {
   RefetchFn,
   FetchPolicy
 } from "./types";
+import type { Disposable } from 'relay-runtime';
```
Markionium, line 13: I see why you have the peerDependency now. 

Is there a way we can use/create a compatible type? That way we do not need to take the whole dependency for just one type.

Markionium: From looking at the types in here [@types/relay-runtime](https://www.npmjs.com/package/@types/relay-runtime?activeTab=code) (Can't direct link, but they're located at `/@types/relay-runtime/lib/util/RelayRuntimeTypes.d.ts`) the types to support this seem really simple :) 

```ts
/**
 * Represents any resource that must be explicitly disposed of. The most common
 * use-case is as a return value for subscriptions, where calling `dispose()`
 * would cancel the subscription.
 */
export interface Disposable {
    dispose: DisposeFn;
}

export type DisposeFn = () => void;
```
```

**Review by Markionium — COMMENTED** [review]
As a side note, perhaps it's worth to have a "test" where we assign the Relay hooks to these so that we're sure they're always type compatible?

File: packages/nova-react/package.json (hunk 1 above)
Comment (DarynaAkhmedova) [reply packages/nova-react/package.json]: No, this dependency is removed :)

File: packages/nova-react/src/graphql/hooks.ts (hunk 2 above)
Comment (DarynaAkhmedova) [reply packages/nova-react/src/graphql/hooks.ts]: Thanks for the suggestion! Code is updated :)

File: packages/nova-react/src/graphql/hooks.test.tsx [hunk 3]
```diff
@@ -14,13 +14,19 @@ import {
...
 import type { GraphQLTaggedNode } from "./taggedNode";
 import type { FragmentRefs } from "./types";
+import type { useMutation as relayUseMutation, GraphQLTaggedNode as RelayGraphQLTaggedNode } from "react-relay";
 
 type IsNotNull<T> = null extends T ? false : true;
 type IsNotUndefined<T> = undefined extends T ? false : true;
 
+// This is used to verify type compatibility with Relay useMutation hook
+declare function useMutationRelay(mutation: RelayGraphQLTaggedNode): ReturnType<typeof useMutation>;
+export const typeVerification: typeof relayUseMutation = useMutationRelay;
```
Markionium, line 28: This is nice :D

packages/nova-react/src/graphql/hooks.ts
```diff
@@ -324,22 +326,27 @@ export function useSubscription<TSubscriptionPayload extends OperationType>(
...
+  onError?: (error: Error) => void;
+}
+
+interface MutationCommitterOptions<TMutationPayload extends OperationType> extends MutationCommitterOptionsBase<TMutationPayload> {
+  onCompleted?: ((response: TMutationPayload["response"], errors: PayloadError[] | null) => void | null) | undefined;
+}
+
+interface MutationCommitterOptions_deprecated<TMutationPayload extends OperationType> extends MutationCommitterOptionsBase<TMutationPayload> {
   onCompleted?: (response: TMutationPayload["response"]) => void;
 }
```
Markionium, line 345: I'm not sure on the inheritance vs just duplicating the type for the options. I'd probably say we should have two? 

My thinking is that when adding more Relay specific options you'd need to make sure to add them to the Options one and not the Base (as that would also change the deprecated one. The other thing is when removing the deprecated one, we could also just remove the type for it. Then we would be left with just the "new" one, instead of the new one and the base (which we'd then probably merge back?)

That said, I'm ok with it if you like to keep it like this.

packages/nova-types/src/nova-graphql.interface.ts
```diff
@@ -64,15 +63,51 @@ export interface NovaGraphQL<GraphQLDocument = any> {
...
+export type Disposable = {
+  dispose(): void;
+};
+
+export interface OperationType {
+  readonly variables: { [name: string]: any };
+  readonly context?: { [name: string]: any };
+  readonly response: any;
+  readonly rawResponse?: unknown;
+}
```
Markionium, line 101: I assume these are also this loosely typed in the Relay typings?

Markionium (dismissed):

packages/nova-types/src/nova-graphql.interface.ts
DarynaAkhmedova: In Relay they have "any" for variables, but "unknown" for the rest. I updated it to be unknown for all since it will be a bit better.

packages/nova-react/src/graphql/hooks.ts
DarynaAkhmedova: Sure, I can keep two of them

packages/examples/src/Feedback/Feedback.tsx
```diff
@@ -1,4 +1,4 @@
-import { graphql, useFragment, useMutation } from "@nova/react";
+import { graphql, useFragment, useMutation_deprecated } from "@nova/react";
```
sjwilczynski, line 1: Let's not use deprecated functions in examples - they should be as up to date as possible

sjwilczynski: Would that mean we need to update graphitation first? If so I would be in favor of that as otherwise nova bump in apollo based hosts will require search and replace and TBH I am not even sure how would we handle that in people app

DarynaAkhmedova: Yes, that would mean that we need to update graphitation first. We discussed it with Mark, and the initial plan was to start with this change. But we can revisit the initial decision

PR #118, opened 2024-10-01
fix(nova-react-test-utils): fix issue with store being shared between stories

sjwilczynski:
As we discovered when integrating new decorator in 1JS, there is a problem with passing relay environment config to the decorator. The issue is that when one used as in docs `getNovaDecorator(schema, { store: createStores() })` that would cause a store to be shared between multiple stories as `createStore` function was executed once for the whole decorator. This means that we cannot just pass relay env options directly instead we change the code to pass function that gets the options.

Tested using storybook and verifying each has unique data in store and results of resolvers don't not slip over. Theoretically it is a breaking change but as the option was not yet widely used, we mark it as minor.

Special thanks to @jakobkhansen for reporting and creating repro of the issue in https://github.com/jakobkhansen/nova-facade/tree/user/jakobhansen/sharedstore_repro

### DIFF
--- a/packages/examples/src/relay/Feedback/Feedback.stories.tsx
+++ b/packages/examples/src/relay/Feedback/Feedback.stories.tsx
@@ -15,14 +15,21 @@ import type { FeedbackStoryQuery } from "./__generated__/FeedbackStoryQuery.grap
 import { getSchema } from "../../testing-utils/getSchema";
 import * as React from "react";
 import type { events } from "../../events/events";
+import { RecordSource, Store } from "relay-runtime";
 
 const schema = getSchema();
 
 const MockPayloadGenerator = new PayloadGenerator(schema);
 
+const novaDecorator = getNovaDecorator(schema, {
+  getEnvironmentOptions: () => ({
+    store: new Store(new RecordSource()),
+  }),
+});
+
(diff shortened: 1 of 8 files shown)


packages/examples/src/relay/Feedback/Feedback.stories.tsx
```diff
@@ -15,14 +15,21 @@ import type { FeedbackStoryQuery } from "./__generated__/FeedbackStoryQuery.grap
...
 
 const schema = getSchema();
 
 const MockPayloadGenerator = new PayloadGenerator(schema);
 
+const novaDecorator = getNovaDecorator(schema, {
+  getEnvironmentOptions: () => ({
+    store: new Store(new RecordSource()),
+  }),
+});
```
jakobkhansen, line 28: Is it intentional to keep this example around? Don't need to have it since Relay will create the store if not given, but I could see it as an example of how to override config.

sjwilczynski: I wanted to keep it as a check for having some stories with no environment and some with custom env. That way we have both cases verified

Markionium (approved):

PR #119, opened 2024-10-01
chore(nova-react-test-utils): refactor for better dependency split

sjwilczynski:
While making two versions of test utils for relay and apollo we realized it's good to separate deps and use optional peer dependencies to make consumers only install one set of deps (for apollo or relay not both). We separated most but in `shared/nova-mock-environment` we kept reference to both `@graphitation/apollo-mock-client` and `relay-test-utils`. In scope of this PR we refactor code to separate these deps out to not be in `shared` folder but only in client specific ones.

Additionally, we change the export map for local development (to refer to src files) to avoid need to build to run tests/stories locally

### DIFF
--- a/packages/nova-react-test-utils/src/shared/shared-utils.ts
+++ b/packages/nova-react-test-utils/src/shared/shared-utils.ts
@@ -4,6 +4,8 @@ import type { makeDecorator } from "@storybook/preview-api";
 
 export type Variant = "apollo" | "relay";
 
+export type Environment = "test" | "storybook";
+
 export function defaultBubble(eventWrapper: EventWrapper): Promise<void> {
   const eventData =
     typeof eventWrapper.event.data === "function"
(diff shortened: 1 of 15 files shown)


packages/nova-react-test-utils/src/shared/shared-utils.ts
```diff
@@ -4,6 +4,8 @@ import type { makeDecorator } from "@storybook/preview-api";
 
 export type Variant = "apollo" | "relay";
 
+export type Environment = "test" | "storybook";
```
Markionium, line 7: Is it confusing to call this `Environment`?  I'm not sure exactly what would be better though.

It's perhaps fine as it's not really used directly.

Markionium (approved):

packages/nova-react-test-utils/src/shared/shared-utils.ts
sjwilczynski: How about TestingEnvironmentType?

sjwilczynski: Did change to something more explicit

PR #120, opened 2024-10-07
fix(nova-react-test-utils): replace default imports with * imports

sjwilczynski:
A lot of packages in 1JS need to update their tsconfig when integrating new test utils, due to default imports on `React` and `invariant`. Instead of forcing the change, we can adjust imports inside the package and remove the need of setting `esModuleInterop`. As verification, we removed that from the tsconfig in the repo itself


Markionium (approved):

alloy (approved):
Thanks!

PR #121, opened 2024-10-08
fix(nova/react): bring back default import for invariant

sjwilczynski:
In #120 we replace all `import React as "react"` and all `import invariant from "invariant"` to be * imports to mitigate issues due to failures in 1JS when seeing `import React as "react"` even in ES modules. However, there were no complaints about invariant and we did change anyway. Turns out that wasn't a good idea because then when running jest in 1JS (or anywhere else were we use /lib files from facade) we get `Error: Uncaught [TypeError: invariant is not a function]` as `invariant` really has a default export which is a single function and then 
image
is missing .default to work :(

So to at least avoid having to set `esModuleInterop` outside of Jest, we bring back the default import for invariant but set `esModuleInterop` for tests only (which is also set in 1JS and TMP in default jest configs)

### DIFF
--- a/change/@nova-react-7777c2ef-6eb5-4013-bbf2-3d51bb071261.json
+++ b/change/@nova-react-7777c2ef-6eb5-4013-bbf2-3d51bb071261.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "bring back default import for invariant",
+  "packageName": "@nova/react",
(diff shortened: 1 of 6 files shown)


Markionium:
Should we look into fixing this in `invariant`?

sjwilczynski:
TBH I am not feeling like trying to push an update to 7 year old package 
image

ansemb:
Yes, let's not do that 😄

PR #127, opened 2024-10-18
fix(nova-react-test-utils): bump relay, relax generate types in relay case, add example and docs for working with client extensions

sjwilczynski:
In scope of this PR we:
* bump relay to ^18
* bump @graphitation packages to latest
* relax type for `generateFunction` so that one can use relay's `MockPayloadGenerator` without type issues (which was the case in 1JS)
* add docs how to work around client extensions not working in current default generator, test it in examples repo


packages/examples/package.json
```diff
@@ -24,17 +24,17 @@
     "graphql": "^15.5.0",
     "react": "^18.3.1",
     "react-dom": "^18.3.1",
-    "react-relay": "^17.0.0",
-    "relay-runtime": "^17.0.0"
+    "react-relay": "^18.0.0",
```
ansemb, line 27: Why not `^18.1.0`?

sjwilczynski: No specific reason, this is internal examples (resolved to 18.1 in lockfile anyway) so there is not real difference for consumers

packages/nova-react-test-utils/README.md
```diff
@@ -277,7 +277,60 @@ const meta = {
...
+import { getNovaDecorator } from "@nova/react-test-utils/relay";
+
+const novaDecorator = getNovaDecorator(schema, {
+  generateFunction: (operation, mockResolvers) => {
+    const result = MockPayloadGenerator.generateWithDefer(
+      operation,
+      mockResolvers ?? null,
+      {
+        mockClientData: true,
+        generateDeferredPayload: true, // this makes sure mock data is array of deferred payloads, not a single response
```
ansemb, line 326: Wouldn't people want this for most use cases? I'm thinking it would be nice to make this the default behavior.

Could we expose a default `generateFunction` from `@nova/react-test-utils/relay` that does exactly this?

ansemb (approved):

packages/nova-react-test-utils/README.md
ansemb: > Even if you check test cases in relay-test-utils a lot of snapshots just don't make sense

Probably our brother GPT that made them :upside_down_face: 

Yes, let's not do this yet.

kerrynf (approved):

PR #129, opened 2024-11-26
feat(nova-react-test-utils): support Storybook 8

sjwilczynski:
As reported some types used in the decorator are no longer available in v8, failing the build when using new version of Storybook in 1JS. In scope of this PR we:
* replace `ComposedStoryPlayContext` with `Partial<PlayFunctionContext>` (we don't move to `StoryContext` yet as that is not containing info about play function in v7)
* migrate examples to local dev to start using v8


packages/examples/.storybook/main.ts
```diff
@@ -48,3 +37,7 @@ const config: StorybookConfig = {
   },
 };
 export default config;
+
+function getAbsolutePath(value: string): string {
+  return dirname(require.resolve(join(value, "package.json")));
```
Markionium, line 42: Given we're using absolute directory I'm assuming that means config paths support backslash on Windows?

Markionium (approved):

packages/examples/.storybook/main.ts
sjwilczynski: Yeah, I am actually having facade checked out on my windows machine :)

PR #130, opened 2025-01-07
docs(nova/react): update readme to mention difference between bubble and generateEvent

sjwilczynski:


packages/nova-react/README.md
```diff
@@ -82,8 +82,8 @@ These events are published in an independent package so that they can be easily
 
 ### Primary Use Cases for Events
 
-- Bubbling a button click that should perform some sort of navigation or external action, like opening a modal on host app side
-- Bubbling an internal action that needs to be logged
+- Bubbling a button click that should perform some sort of navigation or external action, like opening a modal on host app side - in such cases one should rely on `bubble` method, requiring dev to pass related React (click, keyboard) event
```
Markionium, line 85: Would it make sense to add some additional guidance like.

`bubble` should always be used when a React Event is available? Which would be on (almost) all user interactions.

When programatically triggering an event, `generateEvent` can be used.?

Markionium (approved):

packages/nova-react/README.md
sjwilczynski: Updated :)

PR #132, opened 2025-01-14
fix(nova/react): fix eventing for multiple nova event interceptors

carmenberndt:
When there are multiple interceptors, only the last one is called.

The interceptor is reading 'internal' which always refers to the novaEventing set by eventing provider. When setting the context in the interceptor, the same internal is used instead of creating a new one. The internal in the eventing should point to eventing/unmountEventing created by the interceptor and not the root nova eventing.

### DIFF
--- a/packages/nova-react/src/eventing/nova-eventing-provider.test.tsx
+++ b/packages/nova-react/src/eventing/nova-eventing-provider.test.tsx
@@ -533,3 +533,146 @@ describe("NovaEventingInterceptor", () => {
     expect(bubbleCall.event.data()).toBe("addedData");
   });
 });
+
+describe("Multiple NovaEventingInterceptors", () => {
+  const originalError = console.error;
+  beforeEach(() => {
+    jest.clearAllMocks();
+    console.error = originalError;
+  });
+  const realMapper = jest.requireActual(
+    "./react-event-source-mapper",
+  ).mapEventMetadata;
+
+  const mapEventMetadataMock = jest.fn().mockImplementation(realMapper);
+
+  const bubbleMock = jest.fn();
+
+  const parentEventing = {
+    bubble: bubbleMock,
+    generateEvent: bubbleMock,
+  } as unknown as NovaEventing;
+
+  const callbackToBeCalledOnFirstIntercept = jest.fn();
+  const callbackToBeCalledOnSecondIntercept = jest.fn();
+
+  const evenOriginatorToBeInterceptedFirst = "toBeInterceptedFirst";
+  const evenOriginatorToBeInterceptedSecond = "toBeInterceptedSecond";
+
+  const firstInterceptor = (eventWrapper: EventWrapper) => {
+    console.log("received event", eventWrapper.event.originator);
+    if (eventWrapper.event.originator === evenOriginatorToBeInterceptedFirst) {
+      callbackToBeCalledOnFirstIntercept();
+      return Promise.resolve(undefined);
+    } else {
+      return Promise.resolve(eventWrapper);
+    }
+  };
+
+  const secondInterceptor = (eventWrapper: EventWrapper) => {
+    if (eventWrapper.event.originator === evenOriginatorToBeInterceptedSecond) {
+      callbackToBeCalledOnSecondIntercept();
+      return Promise.resolve(undefined);
+    } else {
+      console.log("bubbling up event", eventWrapper.event.originator);
+      return Promise.resolve(eventWrapper);
+    }
+  };
+
+  const ComponentWithTwoEvents = ({
+    name,
+    eventOriginatorToBeIntercepted,
+  }: {
+    name: string;
+    eventOriginatorToBeIntercepted: string;
+  }) => {
+    const eventing = useNovaEventing();
+    const onInterceptClick = (event: React.SyntheticEvent) => {
+      eventing.bubble({
+        reactEvent: event,
+        event: {
+          originator: eventOriginatorToBeIntercepted,
+          type: "TypeToBeIntercepted",
+        },
+      });
+    };
+
+    const onNonInterceptClick = (event: React.SyntheticEvent) => {
+      eventing.bubble({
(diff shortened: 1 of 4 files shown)


carmenberndt:
> please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> Contributor License Agreement

@microsoft-github-policy-service agree company="Microsoft"

packages/nova-react/src/eventing/nova-eventing-provider.tsx
```diff
@@ -241,3 +267,25 @@ const generateEventing =
...
+
+const generateEventing =
+  (
+    eventingRef: React.MutableRefObject<NovaEventing>,
+    interceptorRef?: React.MutableRefObject<
+      (event: EventWrapper) => Promise<EventWrapper | undefined>
+    >,
+  ) =>
+  (): NovaEventing => ({
+    bubble: async (eventWrapper: EventWrapper) => {
```
sjwilczynski, line 279: The code 
```
if (!interceptorRef) {
        return eventingRef.current.bubble(eventWrapper);
      }

      let eventToBubble: EventWrapper | undefined = eventWrapper;
      eventToBubble = await interceptorRef.current(eventWrapper);

      return eventToBubble
        ? eventingRef.current.bubble(eventToBubble)
        : Promise.resolve();
```

repeats in 3 places but it could be shared by extracting a function `getBubble(eventingRef, interceptorRef)`. Than instead of create a confusing pair of two functions with similar name, we can just use that function directly in interceptor inlining the creation of lower level eventings

packages/nova-react/src/eventing/nova-eventing-provider.test.tsx
```diff
@@ -533,3 +533,146 @@ describe("NovaEventingInterceptor", () => {
...
+
+  it("bubbles the event when interceptor returns the event", async () => {
+    const { getByText } = render(<MultipleInterceptorsTestComponent />);
+    const button = getByText(
+      "toBeInterceptedInFirstInterceptor: Fire event without intercept",
+    );
+    button.click();
+    expect(callbackToBeCalledOnSecondIntercept).not.toHaveBeenCalled();
+    expect(callbackToBeCalledOnFirstIntercept).not.toHaveBeenCalled();
+    await waitFor(() => expect(bubbleMock).toHaveBeenCalled());
```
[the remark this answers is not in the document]
sjwilczynski: But all in all, great set of tests :)

packages/nova-react/src/eventing/nova-eventing-provider.tsx
carmenberndt: good point, updated it

alloy:
I have a hard time understanding the semantics. Are we now able to selectively cancel/handle certain subset of events? can you please update/add the jsdoc comments to explain this?

carmenberndt:
> I have a hard time understanding the semantics. Are we now able to selectively cancel/handle certain subset of events? can you please update/add the jsdoc comments to explain this?

The semantics don't change with this bug fix. The event handling with the interceptor can already handle a certain subset of events. Before this PR there was just a bug that only allowed one NovaEventInterceptor in the tree.

packages/nova-react/src/eventing/nova-eventing-provider.test.tsx
```diff
@@ -533,3 +533,147 @@ describe("NovaEventingInterceptor", () => {
...
+  } as unknown as NovaEventing;
+
+  const callbackToBeCalledOnFirstIntercept = jest.fn();
+  const callbackToBeCalledOnSecondIntercept = jest.fn();
+
+  const evenOriginatorToBeInterceptedFirst = "toBeInterceptedFirst";
+  const evenOriginatorToBeInterceptedSecond = "toBeInterceptedSecond";
+
+  const firstInterceptor = (eventWrapper: EventWrapper) => {
+    console.log("received event", eventWrapper.event.originator);
```
Markionium, line 563: Is this left in on purpose?

packages/nova-react/src/eventing/nova-eventing-provider.test.tsx
```diff
@@ -533,3 +533,147 @@ describe("NovaEventingInterceptor", () => {
...
+      return Promise.resolve(eventWrapper);
+    }
+  };
+
+  const secondInterceptor = (eventWrapper: EventWrapper) => {
+    if (eventWrapper.event.originator === evenOriginatorToBeInterceptedSecond) {
+      callbackToBeCalledOnSecondIntercept();
+      return Promise.resolve(undefined);
+    } else {
+      console.log("bubbling up event", eventWrapper.event.originator);
```
Markionium, line 577: Same for this.

Markionium (approved):

packages/nova-react/src/eventing/nova-eventing-provider.test.tsx
carmenberndt: no, removed it

packages/nova-react/src/eventing/nova-eventing-provider.test.tsx
carmenberndt: same here

packages/nova-react/src/eventing/nova-eventing-provider.tsx
```diff
@@ -142,36 +154,63 @@ export const NovaEventingInterceptor: React.FunctionComponent<
...
+      },
+      generateEvent: getGenerateEvent(
+        rootInternal.unmountEventingRef,
+        interceptorRef,
+      ),
+    }),
     [],
   );
 
+  // Internal should point to eventing/unmountEventing created by the interceptor
```
sjwilczynski, line 197: How about `Internal should point to eventing/unmountEventing created by the interceptor, so that we can nest arbitrary numbers of interceptors`?

sjwilczynski (approved):

PR #136, opened 2025-01-16
fix: follow rules of hooks

carmenberndt:
The function creating internal object contains hooks, so it should not be called inside useMemo.

### DIFF
--- a/packages/nova-react/src/eventing/nova-eventing-provider.tsx
+++ b/packages/nova-react/src/eventing/nova-eventing-provider.tsx
@@ -195,11 +195,21 @@ export const NovaEventingInterceptor: React.FunctionComponent<
   );
 
   // Internal should point to eventing/unmountEventing created by the interceptor, so that we can nest arbitrary numbers of interceptors
-  const internal: InternalEventingContext =
-    useCreateInternalEventingContextPointingToInterceptor(
-      rootInternal,
-      interceptorRef,
-    );
+  const { eventing, unmountEventing } = React.useMemo(
+    () =>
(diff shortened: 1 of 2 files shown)


spushkarr (approved):

packages/nova-react/src/eventing/nova-eventing-provider.tsx
```diff
@@ -195,14 +195,11 @@ export const NovaEventingInterceptor: React.FunctionComponent<
   );
 
   // Internal should point to eventing/unmountEventing created by the interceptor, so that we can nest arbitrary numbers of interceptors
-  const internal: InternalEventingContext = React.useMemo(
```
alloy, line 198: I assume the memo was there for a reason, no? But no tests that were covering its necessity, or are there tests for unwanted re-renders and we're still good?

packages/nova-react/src/eventing/nova-eventing-provider.tsx
```diff
@@ -195,13 +195,20 @@ export const NovaEventingInterceptor: React.FunctionComponent<
...
+  );
+  const eventingRef = React.useRef(eventing);
+  const unmountEventingRef = React.useRef(unmountEventing);
+  const internal: InternalEventingContext = React.useMemo(
+    () => ({
+      ...rootInternal,
+      eventingRef,
+      unmountEventingRef,
+    }),
+    [],
```
sjwilczynski, line 211: Should `rootInternal` be a dependency here?

sjwilczynski (approved):

alloy:
Meta observation, are we missing react lint rules here?

PR #139, opened 2025-01-28
[examples] Fix relay-compiler build step

alloy:
Extensions needs to be a dir, not a file


packages/examples/relay.config.json
```diff
@@ -5,10 +5,8 @@
   "projects": {
     "examples": {
       "language": "typescript",
-      "schema": "src/schema.graphql",
-      "schemaExtensions": [
-        "src/clientExtensions.graphql"
-      ]
+      "schema": "data/schema.graphql",
+      "schemaExtensions": ["data/client-extensions"]
```
alloy, line 9: This is the meat of the change

sjwilczynski: How I love the bugs related to usage of different platforms <3 I assume this time it is Mac specific issue

sjwilczynski: We need the changefile but otherwise LGTM

sjwilczynski (approved):

PR #140, opened 2025-01-28
Add localization support

Markionium:
This PR adds a Localization context, specifically to enable Nova components to consume localized strings with placeholders in a typesafe manner.

For more details around how this work see the updated README.md in this PR.


packages/nova-react/README.md
```diff
@@ -197,3 +197,135 @@ const MyComponentWrapper = () => {
...
+    viewData
+  )
+
+  const format = useFormat();
+
+  return <div>{format(greeting, { name })}</div>;
+};
+```
+
+### Providing the `useFormat` hook implementation
```
sjwilczynski, line 303: Apart from update in readme, could we also update `Feedback` in `examples` folder to use localization in some way?

packages/nova-types/src/nova-localization.ts
```diff
@@ -0,0 +1,29 @@
...
+    : P
+  : never;
+
+type UseFormat = () => (
+  value: unknown,
+  values: Record<string, string | number>,
+) => string;
+
+export type NovaLocalization = {
+  useFormat: UseFormat;
```
kerrynf, line 20: This isn't a hook right, just a normal function? I think this should just be called format?

We don't have any tie to React on the types side right now, and I'd rather not confuse things with a use prefix here... :)

kerrynf: Especially based on the usage here, this one should be just format...

image

packages/nova-react/README.md
```diff
@@ -197,3 +197,135 @@ const MyComponentWrapper = () => {
 The `NovaEventingInterceptor` will intercept the event and if you can check it's properties to decide if is should be acted upon. If from `intercept` promise resolving to undefined is returned the event will not be passed to eventing higher up the tree. However, if to process the event further, one should return a promise resolving to the `eventWrapper` object. That also gives a possibility to alter the event and still pass it further up.
 
 You can nest as many interceptors as you need to either handle or pass events further up.
+
+## Localization
+
+> Note for this to work you will need to use the Microsoft internal fork of the Relay compiler that includes the added functionality.
```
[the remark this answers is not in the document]
alloy: Yes, we discussed that this would only be an issue if the host would be compiling _and_ type-checking the nova components.

packages/nova-react/README.md
```diff
@@ -197,3 +197,135 @@ const MyComponentWrapper = () => {
...
+
+Localized strings are provided to components through GraphQL fragments. For simple strings these will be typed as `string` and can be used directly. For strings that include placeholders Relay the compiler will type them as `StringWithPlaceholders` from `@nova/react`.
```
alloy, line 205:
```suggestion
Localized strings are provided to components through GraphQL fragments. For simple strings these will be typed as `string` and can be used directly. For strings that include placeholders the Relay compiler will type them as `StringWithPlaceholders` from `@nova/react`.
```

packages/nova-react/src/localization/nova-localization.tsx
```diff
@@ -0,0 +1,38 @@
...
+import type { NovaLocalization, FormatFn } from "@nova/types";
+import invariant from "invariant";
+
+const NovaLocalizationContext = React.createContext<NovaLocalization | null>(
+  null,
+);
+
+type NovaLocalizationProviderProps = {
+  localization: NovaLocalization;
+  children?: React.ReactNode | undefined;
```
alloy, line 11: Nits: You can use the `React.PropsWithChildren` typing, but also do we even need to support `undefined` here with a provider?

Anyways, not in any way blocking.

alloy:
Some minor comments, but otherwise LGTM 👍

packages/nova-react/README.md
Markionium: We could add localization without placeholders but for any strings with placeholders we'll need compiler support.

packages/nova-react/README.md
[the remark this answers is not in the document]
Markionium: I'm not sure we're all aligned on this? Eloy is right that the hosts can consume these components if they don't type check the Nova components. 

But what I think @kerrynf is asking is if authors could still author components without having compiler support. I asked if we should support that scenario and I think we said we would not support that. 

But perhaps we want to reconsider?

packages/nova-types/src/nova-localization.ts
Markionium: I guess technically you're right that it is not a hook, but it could be which makes it much more flexible.

If you look at how it's used inside the implementation, it's consumed from the context and then called to return a function.

```ts
export function useFormat(): FormatFn {
  return useLocalization().useFormat();
}
```

In my mind this is similar to how we treat the `useFragment` hook on the GraphQL context. Technically that doesn't need to be a hook either if you have a very simple implementation. But if you use `relay` you want it to be a hook.

If you consider how to integrate this with your hosts localization library, which you might need to access through a hook. Being able to use hooks would be pretty helpful

```tsx
import { NovaLocalizationProvider, type NovaLocalization } from "@nova/react";

const localization: NovaLocalization = {
  useFormat: () =>
    const { t } = useTranslation();

    return (string: string, values: Record<string, string>) =>  t(string, values);
  }
}

const App = () => {
  return (
    <NovaLocalizationProvider localization={localization}>
      <OtherComponents />
    </NovaLocalizationProvider>
  );
};
```

packages/nova-react/src/localization/nova-localization.tsx
Markionium: I'll fix that :D

packages/nova-types/src/nova-localization.ts
Markionium: @kerrynf would be good if you could review this. I'm fine to limit it to a pure function, but I think we'd restrict the implementors more than necessary.

packages/nova-react/README.md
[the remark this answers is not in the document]
Markionium: I adjusted the types so it supports both. 

I also added some examples to the `examples` folder that work without compiler support (and thus without placeholder type safety).

@alloy I know this isn't exactly what we agreed at first but it's probably easier on adoption/consumption across repos.

packages/nova-react/README.md
Markionium: Done for both and reworked the typings a bit to allow it without type safe placeholders.

packages/examples/src/relay/pure-relay/Feedback.stories.tsx
```diff
@@ -177,7 +180,9 @@ export const LikeFailure: Story = {
         { mockClientData: true },
       );
     });
-    const likeButton = await container.findByRole("button", { name: "Like" });
+    const likeButton = await container.findByRole("button", {
+      name: '<mock-value-for-field-"likeLabel">', // No mock resolvers available here, so we use the generated value
```
sjwilczynski, line 184: So let's add them, I think just specifying resolver for ViewData in line 178, would be enought

packages/examples/src/relay/Feedback/Feedback.tsx
```diff
@@ -1,4 +1,4 @@
-import { graphql, useFragment, useMutation } from "@nova/react";
+import { graphql, useFormat, useFragment, useMutation } from "@nova/react";
```
sjwilczynski, line 1: Did we change that component or just added unused import? I would be good to keep pure-relay and relay in sync

packages/nova-react-test-utils/src/relay/test-utils.tsx
```diff
@@ -9,6 +9,8 @@ import { novaGraphql } from "./nova-relay-graphql";
 import { createMockEnvironment } from "relay-test-utils";
 import { RelayEnvironmentProvider } from "react-relay";
 import type { NovaMockEnvironment } from "./nova-mock-environment";
+import { NovaLocalizationProvider } from "@nova/react";
```
sjwilczynski, line 12: Is this import used somewhere?

sjwilczynski (approved):

packages/examples/src/relay/Feedback/Feedback.tsx
Markionium: Ah good catch, it's just left over import from when I was trying something.

packages/nova-react-test-utils/src/relay/test-utils.tsx
Markionium: No, I imported it, but the auto import wasn't removed.

packages/examples/src/relay/pure-relay/Feedback.stories.tsx
Markionium: Perfect! :D I tried to add them to this place first, but that's not allowed when `enableQueuedMockResolvers` is false.

```ts
 parameters: {
    novaEnvironment: {
      enableQueuedMockResolvers: false,
    },
  } satisfies WithNovaEnvironment<FeedbackStoryRelayQuery, TypeMap>,
  ```
  
  I somehow missed the resolvers further down in the test. Thanks!

File: packages/examples/src/relay/Feedback/Feedback.tsx (hunk 7 above)
Comment (Markionium) [reply packages/examples/src/relay/Feedback/Feedback.tsx:1]: @sjwilczynski the problem with the `relay` one is that it uses Graphitation payload generator which doesn't support  client side fields (yet).

Only the `pure-relay` example has viewData today.

File: packages/nova-types/src/nova-localization.ts [hunk 9]
```diff
@@ -0,0 +1,29 @@
+const LocalizedString = Symbol("LocalizedString");
+
+export type StringWithPlaceholders<T extends Record<string, string | number>> =
```
prprabhu-ms, line 3: Let's add docstring for this type. In particular, what are the semantics of the key and value in the `Record` type in the type parameter constraint?

packages/nova-types/src/nova-localization.ts
```diff
@@ -0,0 +1,29 @@
...
+  {
+    [LocalizedString]: T;
```
prprabhu-ms, line 5: Is this intentional that the symbol `LocalizedString` is not exported? A way to have a newtype to hide the actual, irrleevant, way this type is constructed?

packages/nova-types/src/nova-localization.ts
```diff
@@ -0,0 +1,29 @@
...
+  };
+
+export type Placeholders<T> = T extends StringWithPlaceholders<infer P>
+  ? keyof P extends never // Make sure that the placeholder object is not empty
+    ? never
+    : P
+  : never;
+
+type UseFormat = () => (
```
prprabhu-ms, line 14: Please add a docstring explaining the semantics of the arguments here. Also, let's try to name them better than `value` and `values`. Currently, reader can't saay how these two differ.

prprabhu-ms: Also, I think you intended to export `UserFormat` as it is used in an exported type below.

packages/nova-types/src/nova-localization.ts
```diff
@@ -0,0 +1,29 @@
...
+  value: unknown,
+  values: Record<string, string | number>,
+) => string;
+
+export type NovaLocalization = {
+  useFormat: UseFormat;
+};
+
+export type FormatFn = <
```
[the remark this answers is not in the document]
prprabhu-ms: If this is needed, please add a docstring. All exported types need a docstring because readers should be able to stop at the interface and not have to dig into the implementation to understand what to do with the type.

packages/examples/src/relay/pure-relay/Feedback.stories.tsx
```diff
@@ -173,11 +176,19 @@ export const LikeFailure: Story = {
         operation,
         {
           Feedback: () => sampleFeedback,
+          ViewData: () => ({
+            viewDataField: "View data field",
+            likeLabel: "Like",
+            unlikeLabel: "Unlike",
+            numberOfLikesLabel: "You have {count} likes",
+          }),
```
alloy, line 184: Nit: I think it would speak more to the imagination if the example was using different languages for the string definition vs the runtime value [for the current locale]. Not a biggie, though.

alloy (approved):
@Markionium Looks great, thanks for all the excellent work 👏 

Agreed with @prprabhu-ms to ensure we have docs on the exports, where possible/needed.

packages/nova-types/src/nova-localization.ts
Markionium: Yes, it makes it more difficult for folks to work around it by manually creating something. 

It would still be possible to cast though, so that workaround is always available

packages/nova-types/src/nova-localization.ts
Markionium: In most cases you will not need to use this type the compiler will generate the usage, in which case the type parameter would be pretty clear. That said, I'll add some words on how it's used.

packages/nova-types/src/nova-localization.ts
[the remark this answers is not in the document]
Markionium: It's left over from before, I'll remove this.

PR #141, opened 2025-01-29
Add example Relay provider and export relay Feedback example component

alloy:
Per discussion with the PowerAutomate folks, they are wanting to integrate some Nova components. We'll be making the Feedback example component available as a canary Nova component to test the host's Nova facade setup.

This first PR only implements a Relay GraphQL provider. Adding an eventing and commanding provider should be straightforward.

⚠️  The changes in this PR are untested, as I have no bandwidth to setup a test host application to test these in. Instead, let's publish a new release and ask the PowerAutomate folks to test it, fix it, expand it.


alloy:
@sjwilczynski Unless you see any hard blockers, can you please merge this as soon as you have reviewed it? I would like to have a package published by tomorrow morning pacific time.

packages/examples/src/relay/NovaExampleRelayGraphQLProvider/useRelayMutationWithApolloAPI.ts
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * Hook that wraps original Relay useMutation hook and returns mutation committer,
+ * that returns a promise. This is required for alignment with Apollo client
+ * useMutation hook API for smoother migration of production code to Relay.
+ * @param mutation Mutation document.
+ * @returns Array containing mutation committer function and a flag
+ * indicating whether mutation is "in flight".
+ */
+export function useRelayMutationWithApolloAPI<
```
sjwilczynski, line 25: With this change we start having two different implementation of Nova relay based environment, the other one exists in `packages\nova-react-test-utils\src\relay\nova-relay-graphql.ts`. We shouldn't, it's going to be confusing as hell why certain things work in Storybook differently than in imported env from examples package.

In 1JS (and in Outlook) nova env based on relay is using suspending `useLazyLoadQuery` and we accepted the consequence that Nova interface is not exactly followed (as we never return error) and TBH I am now surprised Teams went in different direction. Seems we need to discuss and align on this in a forum.

For now I will approve and merge the PR, to allow you to share this with folks, but without this need I would block it until we aligned, so please follow up on it

packages/examples/src/relay/NovaExampleRelayGraphQLProvider/useRelayLazyLoadQueryWithApolloAPI.ts
```diff
@@ -0,0 +1,30 @@
...
+import { useQuery } from "relay-hooks";
+import type {
+  FetchPolicy,
+  GraphQLTaggedNode,
+  OperationType,
+} from "relay-runtime";
+
+/**
+ * Hook that wraps useQuery hook from "relay-hooks" library and returns a Nova API compatible response.
+ * We can't use original Relay useLazyLoadQuery, since its implementation is heavily relies on Suspense.
```
sjwilczynski, line 16:
```suggestion
 * We can't use original Relay useLazyLoadQuery, since its implementation heavily relies on Suspense.
```

sjwilczynski (approved):

packages/examples/src/relay/NovaExampleRelayGraphQLProvider/useRelayMutationWithApolloAPI.ts
sergey-stoyan: Sure, let's discuss during the next sync

PR #142, opened 2025-02-11
fix(pipeline): line ending fix

pavelglac:
The release pipeline is failing due to differences in the example package caused by inconsistent line endings in the files. We should not have committed the generated documents since they are automatically generated with each `yarn install`.


sjwilczynski:
TBH we don't need them, I didn't check them on purpose, I agree with removing them

sjwilczynski (approved):

PR #143, opened 2025-02-11
fix(example): remove unused document

pavelglac:
This PR removing unused file which causing the diff


pavelglac:
Yes indeed, thanks.

PR #148, opened 2025-03-24
feat(nova/react-tets-utils): throw an error if user has client extensions only query in relay variant

sjwilczynski:
Multiple users in 1JS complained that it is hard to figure out why mock data is not being generated in certain cases and specified resolvers are not even executed. Turns out it was because in Relay client extensions only queries don't even reach "server" resolvers so there is nothing to generate, check facebook/relay#4927.

To make it less confusing for our devs, until we have either a fix in relay or graphitation payload generator supporting client extensions, let's throw an informative error, to let devs know what they should do.

### DIFF
--- a/packages/examples/src/relay/pure-relay/ViewDataOnly.test.tsx
+++ b/packages/examples/src/relay/pure-relay/ViewDataOnly.test.tsx
@@ -13,7 +13,7 @@ describe("ViewDataOnly", () => {
     expect(screen.getByText(/This is a view data field/i)).toBeInTheDocument();
   });
 
-  it("throws an expcetion when only client fields are selected in test query", () => {
+  it("throws an exception when only client fields are selected in test query", () => {
     expect(() => {
       render(<ViewDataOnlyStory />);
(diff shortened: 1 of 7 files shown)


packages/examples/src/relay/pure-relay/ViewDataOnly.test.tsx
```diff
@@ -0,0 +1,23 @@
...
+const { ViewDataOnlyStory, ViewDataOnlyWithServerFieldSelected } =
+  composeStories(stories);
+
+describe("ViewDataOnly", () => {
+  it("renders ViewDataOnly component", () => {
+    render(<ViewDataOnlyWithServerFieldSelected />);
+    expect(screen.getByText(/This is a view data field/i)).toBeInTheDocument();
+  });
+
+  it("throws an expcetion when only client fields are selected in test query", () => {
```
kerrynf, line 16:
```suggestion
  it("throws an exception when only client fields are selected in test query", () => {
```

PR #150, opened 2025-03-31
feat(examples): migrate to examples to Storybook test addon

sjwilczynski:
The initial plan for Storybook folks to make stories behave as tests, was using [@storybook/test-runner](https://storybook.js.org/docs/writing-tests/test-runner). As useful as it was, the fact that it used jest as orchestrator made things slower than they should be. On top of that the current default for Storybook + React is to use vite instead of webpack just because of perf gains. No surprise then that to make the vite integration even more simplest, they wanted to integrate with vitest, their test runner. Their result of that integration is [test addon](https://storybook.js.org/docs/writing-tests/test-addon). Even tough it is marked as experimental, Storybook folks already announced this as future of their testing workflow. Check how nicely it integrates both with command line and UI:

testAddon

You can run your stories as tests both from UI and from command line without any issues with awesome DX provided by vitest which is significantly faster than jest.

1JS is still on Storybook 7 so might be a bit harder to use it like that there but other repos have the benefit of being able to use latest greatest and (looking at you @kerrynf ) I would recommend start to experiment with it there.

This allows us to migrate majority of jest tests directly into Storybook, only left the ones that really verify something in jest environment. Part of this PR is extracted to #151 for easier review


alloy (approved):
Other than the request to check if source-maps are working correctly, this looks great!

PR #151, opened 2025-03-31
feat(examples): migrate Storybook from webpack to vite

sjwilczynski:
Extracted for clarity from #150.

In scope of this PR we migrate our Storybook from webpack to vite, which is the recommended setup from Storybook folks. It also makes Storybook faster and allows for using test addon, which is the new super awesome way of writing test using Storybook. 

As a prerequisite we migrate apollo based stories to `duct-tape-compiler` as otherwise `babel-plugin-relay` would complain and anyway, there was no real value in keeping the old setup which no team should use.


packages/examples/.storybook/main.ts
```diff
@@ -1,39 +1,27 @@
...
-          {
-            test: /\.(graphql)$/,
-            loader: "@graphql-tools/webpack-loader",
-          },
-        ],
-      },
-    };
+  viteFinal: (config) => {
+    return mergeConfig(config, {
+      plugins: [relayPlugin, graphqlLoader(), commonjs()],
```
ansemb, line 23: why do we need `commonjs()`? meaning do we have downstream component packages that is using `commonjs()` currently?

I kinda understand if it's for backward compatibility, if it's really needed. However, I would prefer not allowing commonjs, as this is not how we want to write code - using commonjs is simply doing it wrong.

ansemb (approved):

packages/examples/.storybook/main.ts
alloy: Relay uses commonjs

alloy (approved):
:shipit:

PR #152, opened 2025-03-31
fix(react-test-utils): be more restrictive when detecting client extension only queries

sjwilczynski:
In #148 we added a logic which would throw exception if dev used client extension only query in relay variant of the decorator as we cannot generate mock data for it (check original PR for details). However, it turned out not fully correct:
* the client extensions only query never reaches a server (so doesn't call mock resolvers) if no server field is added to the query
* the client extensions only query does reach a server and calls resolver if at selection path at least one field is a server field

In scope of this PR, we limit the check to only look for the first case and let decorator behave as before for second one


Markionium:
Should we figure out how we would like to also be able to test this? Do we have many of these cases?

> the client extensions only query never reaches a server (so doesn't call mock resolvers) if no server field is added to the query

sjwilczynski:
> Should we figure out how we would like to also be able to test this? Do we have many of these cases?

As I mentioned in original PR the short term solution is adding server field which is anyway what folks in 1JS have been doing

Long term solution is either for original issue to be fixed in relay mock environment or duct tape generator support client extensions

alloy (approved):
LGTM 👍

PR #156, opened 2025-04-07
chore: move all packages away from jest

sjwilczynski:
This PR is logical continuation of #150 - we in scope of this PR migrate away from using jest in the repository completely. Here are some performance numbers between jest and vitest, for [script doing the benchmark check the gist](https://gist.github.com/sjwilczynski/267db495e45e6fc18e07e73b86f757e7). Pasting here for convenience:

Vitest:

| Package                | Average Time (s) | Std Deviation (s) | Min Time (s)   | Max Time (s)   |
|------------------------|------------------|-------------------|----------------|----------------|
| examples               |            18.96 |              5.26 |          16.84 |          38.57 |
| examples  (with swc)        |            16.10 |              2.46 |          14.70 |          25.22 |
| nova-graphql-compiler  |             2.23 |              0.24 |           2.13 |           3.15 |
| nova-react             |             4.40 |              0.10 |           4.33 |           4.77 |
| nova-react-test-utils  |             2.61 |              0.16 |           2.54 |           3.20 |
---------------------------------------------------------------------------------------------------

Jest (before moving most tests in `examples` to stories):

| Package                | Average Time (s) | Std Deviation (s) | Min Time (s)   | Max Time (s)   |
|------------------------|------------------|-------------------|----------------|----------------|
| examples               |            19.53 |              6.64 |          16.15 |          44.31 |
| nova-graphql-compiler  |             7.66 |              0.64 |           7.39 |          10.04 |
| nova-react             |            16.14 |              1.62 |          15.38 |          22.15 |
| nova-react-test-utils  |             6.92 |              0.57 |           6.60 |           8.97 |
---------------------------------------------------------------------------------------------------

**Notes**: 
* the SWC version is the one currently in the PR.
* the max time refers to case when we just checked out the repo, ran `yarn` and immediately after started running tests

We can see 3-4 times fast perf on smaller packages and ~20% improvement faster in bigger much (with much higher amount of tests, because now all stories are included) and ~45% faster cold start, on newly checked out repo.

Some additional benefits:
* we use [Vitest browser mode](https://vitest.dev/guide/browser/why.html), which means tests are run in headless browser instead of poor simulation like JSDOM in jest, no more cases of something working in browser but not when run in Jest
* `test` verb no longer depends on `build` or `types` as vitest doesn't have limitations like ts-jest
* Vitest has proper support for ESM so we no longer need to worry about complex and counterintuitive jest transforms
* We share a lot of test config between unit tests and stories
* We move away from testing-library in favor of `vitest-browser-react` with similar but a bit nicer API with playwright like capabilities

Some additional changes:
* we now have share vitest config, the other packages extend or reuse
* we simplified tsconfig to have shared options in root and only override on each package level
* **BREAKING CHANGE**: in @nova/react-test-utils we make a breaking change as previously we hardcoded `bubble` and `trigger` to be instances of jest mock for variant of util exposed to unit tests. As we now want to support both, instead of having default mock, we just give an option to pass mock from your testing framework of choice. In 1JS we don't depend on that behavior but in semantic versioning this needs to be a major bump
* We managed to get rid of `vite-plugin-commonjs` (cc @ansemb) by first replacing `require` statement when loading schema and then by removing `vite-plugin-relay` and related Babel plugin by replacing it with `@swc/plugin-relay` to align more with tools used in our outside repositories.


scripts/just.config.ts
```diff
@@ -1,12 +1,5 @@
 import type { EsbuildBuildOptions } from "just-scripts";
-import {
-  tscTask,
-  esbuildTask,
-  jestTask,
-  eslintTask,
-  argv,
-  parallel,
-} from "just-scripts";
+import { tscTask, esbuildTask, eslintTask, parallel } from "just-scripts";
```
Markionium, line 2: Any opinions on if we need/want a `just-scripts` task for vitest?

packages/nova-react/src/localization/nova-localization.test.tsx
```diff
@@ -24,35 +20,27 @@ describe(useFormat, () => {
   it("is able to access the `format` function provided by the provider", () => {
     expect.assertions(3);
 
-    const formatFn = jest.fn();
+    const formatFn = vi.fn();
     const localization: NovaLocalization = {
-      useFormat: jest.fn().mockReturnValue(formatFn),
+      useFormat: vi.fn().mockReturnValue(formatFn),
     };
-
```
Markionium, line 31: No love for new lines? 🤷

packages/nova-react/src/graphql/taggedNode.test.ts
```diff
@@ -1,7 +1,14 @@
+/**
+ * @vitest-environment node
```
Markionium, line 2: For my own curiosity and anyone reading next :) 

https://vitest.dev/guide/environment.html#environments-for-specific-files

packages/nova-react/src/graphql/nova-graphql-provider.test.tsx
```diff
@@ -1,42 +1,32 @@
...
 import * as React from "react";
 import type { NovaGraphQL } from "@nova/types";
-import { render } from "@testing-library/react";
+import { render } from "vitest-browser-react";
+import { describe, it, expect, vi } from "vitest";
 import { NovaGraphQLProvider, useNovaGraphQL } from "./nova-graphql-provider";
 
 describe(useNovaGraphQL, () => {
   it("throws without a provider", () => {
-    expect.assertions(1);
```
Markionium, line 12: I noticed we left these in some of the other tests. Any specific reason why they're removed here?

packages/nova-react-test-utils/src/shared/nova-mock-environment.tsx
```diff
@@ -14,24 +14,16 @@ import {
...
 
-type Commanding<T extends TestingEnvironmentVariant> = T extends "test"
-  ? jest.Mocked<NovaCentralizedCommanding>
-  : NovaCentralizedCommanding;
+type Commanding = NovaCentralizedCommanding;
 
-type Eventing<T extends TestingEnvironmentVariant> = T extends "test"
-  ? jest.Mocked<NovaEventing>
-  : NovaEventing;
+type Eventing = NovaEventing;
```
Markionium, line 22: Probably don't need these type aliases?

packages/examples/src/apollo/Feedback/FeedbackContainer.test.tsx
```diff
@@ -30,24 +34,23 @@ describe("FeedbackContainer", () => {
...
+        const texts = page.getByText(text);
+        expect(texts.elements()).toHaveLength(3);
       },
       { timeout: 3000 },
     );
-    expect(screen.getAllByText(text)[2]).toBeInTheDocument();
   });
 
-  // kept in jest to ensure prepareStoryContextForTest works correctly
+  // kept in unit to ensure prepareStoryContextForTest works correctly
```
Markionium, line 46:
```suggestion
  // kept in unit test to ensure prepareStoryContextForTest works correctly
```

Markionium (approved):
Great stuff! 🥳

packages/nova-react/src/localization/nova-localization.test.tsx
sjwilczynski: Turns out agent doesn't love them when doing automated refactor, can try to bring most of them back tough

packages/nova-react/src/graphql/nova-graphql-provider.test.tsx
sjwilczynski: Will check carefully for these and bring back - agent didn't like them for some reason

scripts/just.config.ts
sjwilczynski: I don't think we need it. In general I am not the biggest fan of this abstraction layer especially that unless configured, it doesn't allow passing more arguments to a command. I personally find it confusing and it doesn't bring value, sharing config is anyway done through just by extending default config in each package.

However, if someone could show me some clear benefits, I am open to changing my mind

Markionium: Agreed with that, I phrased it as a question to not anchor the comment.

PR #157, opened 2025-04-10
Add additional react version

kerrynf:
Adding additional React version in the peer dependencies to see if this helps with integration into React 16 host.


alloy (approved):
Thanks! For posterity, this is for PowerAutomate, we will need to follow up on whether we can actually support v16 for any nova packages.

PR #159, opened 2025-05-05
deps: bump vitest and Storybook

sjwilczynski:
In scope of this PR we bump to latest 8.x Storybook and [vitest 3.2](https://vitest.dev/blog/vitest-3-2.html) which includes some nice performance gains


alloy (approved):
🙌

PR #163, opened 2025-06-05
Add @nova/schema package

alina-zaieva:
Nova defines extra pieces of SDL to allow high-value components to provide extra metadata as a part of their GraphQL schema. This new package will be used to host these SDL declarations, and can be used both for documentation purposes and for GraphQL tooling.

### DIFF
--- a/packages/nova-schema/package.json
+++ b/packages/nova-schema/package.json
@@ -1,0 +1,12 @@
+{
+  "name": "@nova/schema",
+  "version": "0.0.0",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/microsoft/nova-facade.git",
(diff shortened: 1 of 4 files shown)


sjwilczynski:
LGTN

packages/nova-schema/package.json
```diff
@@ -0,0 +1,12 @@
+{
+  "name": "@nova/schema",
+  "version": "0.0.0",
```
Markionium, line 3: Will the change file bump this to 1.0.0?

alina-zaieva: Yes, tested it locally via `beachball bump` 🙂

Markionium (approved):

PR #166, opened 2025-08-01
feat(nova/react): introduce forwardEvent API to interceptor

sjwilczynski:
In 1JS we have a number of cases, where we need to intercept an event (like input submission in one of the agents) and based on that trigger multiple other events (navigation, drawer close, etc). We cannot do that in root event handler as that logic is specific to our agent while the event itself is shared. 

To help us work with that requirements, in scope of this PR, we add another parameter on interceptor called `forwardEvent`, which allows from within the interceptor to forward a new event to next layer. 

We added proper section for it in README, updated tests to include the new usage pattern and added a type for interceptor to be shared. 

<details>

<summary>In case someone is interested this is the plan file used for agent mode to help us build this feature</summary>

# Nova Interceptor Multiple Events Support Plan

## Current State

The current `NovaEventingInterceptor` has a limitation where the interceptor function can only:

- Return a single `EventWrapper` to forward up the chain
- Return `undefined` to prevent the event from bubbling up
- Process only one event at a time

## Problem Statement

From the discussion, there are scenarios where an interceptor needs to:

1. Trigger multiple events based on a single incoming event
2. Forward the original event along with additional generated events
3. Handle user interactions that should result in multiple logged/tracked events

### 1. Updated Interceptor Function Signature

Update the existing interceptor signature to include the `forwardEvent` parameter:

```typescript
interface NovaEventingInterceptorProps {
  interceptor: (
    event: EventWrapper,
    forwardEvent: (event: EventWrapper) => Promise<void>
  ) => Promise<EventWrapper | undefined>;
  children?: React.ReactNode | undefined;
}
```

### 2. Key Design Principles

1. **Simple API**: Single interceptor signature with optional `forwardEvent` usage
2. **Clear Semantics**: The return value still controls the original event, `forwardEvent` is for additional events
3. **Async Support**: Both the interceptor and `forwardEvent` are async to maintain consistency
4. **Type Safety**: Full TypeScript support for the signature
5. **Minimal Breaking Change**: Only adds a parameter that consumers can ignore if not needed

#### Phase 1: Update Interface and Types

- Update `NovaEventingInterceptorProps` interface to include `forwardEvent` parameter
- Update all internal type references

#### Phase 2: Core Logic Updates

- Modify `getBubble` and `getGenerateEvent` functions to provide `forwardEvent` implementation
- Implement `forwardEvent` function that bypasses the current interceptor but respects parent interceptors
- Update internal event handling

#### Phase 3: Update Component Implementation

- Update `NovaEventingInterceptor` component to provide `forwardEvent` implementation to interceptors
- Ensure proper error handling and async behavior

#### Current Usage (needs parameter update)

```typescript
const interceptor = async (event: EventWrapper, forwardEvent: (event: EventWrapper) => Promise<void>) => {
  // Existing interceptors just ignore the forwardEvent parameter
  if (shouldIntercept(event)) {
    return undefined; // block event
  }
  return event; // forward event
};
```

#### New Multiple Events Usage

```typescript
const interceptor = async (event: EventWrapper, forwardEvent: (event: EventWrapper) => Promise<void>) => {
  if (event.event.eventType === "userAction") {
    // Forward additional tracking event
    await forwardEvent({
      event: { eventType: "trackingEvent", originator: "interceptor" },
      source: event.source
    });
    
    // Forward analytics event
    await forwardEvent({
      event: { eventType: "analyticsEvent", data: () => ({ action: "clicked" }) },
      source: event.source
    });
  }
  
  return event; // Still forward original event
};
```

#### A. Updated Interface

- Update the existing `NovaEventingInterceptorProps` interface
- No need for type discrimination or multiple interfaces

#### B. forwardEvent Implementation

- Should bypass current interceptor but respect parent interceptors
- Should use the same eventing reference as the original event
- Should maintain proper async error handling

#### C. Testing Strategy

- Update existing tests to provide the new parameter (even if unused)
- Add new tests for multi-event scenarios
- Test nested interceptors with multiple events
- Test error scenarios (forwardEvent failures, etc.)

### 6. Migration Path

1. **Phase 1**: Update interceptor signature and implementation
2. **Phase 2**: Update existing interceptors to include the new parameter (even if unused)
3. **Phase 3**: Update documentation and examples
4. **Phase 4**: Start using multiple events where beneficial

### 7. Breaking Change Impact

1. **Type Level**: Existing interceptors will have TypeScript errors until they add the parameter
2. **Runtime Level**: Minimal impact - interceptors just need to accept the extra parameter
3. **Migration Effort**: Simple parameter addition to existing interceptor functions

### 8. Potential Edge Cases

1. **Nested Interceptors**: How do forwarded events interact with parent interceptors?
   - **Decision**: Forwarded events should bypass the current interceptor but still go through parent interceptors
2. **Error Handling**: What happens if `forwardEvent` throws?
   - **Decision**: Should not affect the original event processing
3. **Performance**: Multiple async calls in sequence vs parallel
   - **Decision**: Allow parallel forwarding, don't await in sequence unless necessary
4. **Event Ordering**: Should forwarded events maintain any ordering guarantees?
   - **Decision**: No specific ordering guarantees, events are processed as they resolve

### 9. Benefits

- **Simplicity**: Single interface, no type complexity
- **Flexibility**: Interceptors can now handle complex multi-event scenarios
- **Separation of Concerns**: Original event handling vs additional event generation
- **Minimal Breaking Change**: Only requires adding a parameter
- **Type Safety**: Full TypeScript support prevents runtime errors
- **Performance**: Async design allows for efficient event processing

### 10. Implementation Files to Modify

1. `nova-eventing-provider.tsx` - Update interface and core interceptor logic
2. `nova-eventing-provider.test.tsx` - Update tests and add new multi-event test coverage
3. `nova-eventing.interface.ts` - Add new `EventInterceptor` type for reusability
4. Documentation/README updates

### 11. Success Criteria

- ✅ Updated interceptor signature works correctly
- ✅ New `EventInterceptor` type is defined and reused throughout the codebase
- ✅ New multi-event functionality works as specified
- ✅ Performance impact is negligible
- ✅ TypeScript compilation works correctly
- ✅ Existing interceptors can be easily migrated by adding the parameter
- ✅ Documentation clearly explains the usage patterns
</details>

### DIFF
--- a/change/@nova-react-a4bbd969-ef68-4499-8630-d3f1408625d6.json
+++ b/change/@nova-react-a4bbd969-ef68-4499-8630-d3f1408625d6.json
@@ -1,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "add new forwardEvent api",
+  "packageName": "@nova/react",
+  "email": "stwilczy@microsoft.com",
+  "dependentChangeType": "patch"
+}
--- a/change/@nova-react-test-utils-a030f59b-7dc9-4fcc-8f1e-2cc3b29ece8e.json
+++ b/change/@nova-react-test-utils-a030f59b-7dc9-4fcc-8f1e-2cc3b29ece8e.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "use new EventInterceptor type",
+  "packageName": "@nova/react-test-utils",
+  "email": "stwilczy@microsoft.com",
+  "dependentChangeType": "patch"
+}
--- a/change/@nova-types-ab07b1a6-bd11-4a21-8f98-5fa5ebbfc032.json
+++ b/change/@nova-types-ab07b1a6-bd11-4a21-8f98-5fa5ebbfc032.json
@@ -1,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "expose EventInterceptor type",
+  "packageName": "@nova/types",
+  "email": "stwilczy@microsoft.com",
+  "dependentChangeType": "patch"
+}


alloy (approved):
I'm not opposed to this per se, but am wondering why you couldn't just use the existing `bubble` API inside the handler? In the end it seems like that's what it does under the hood anyways, no?

PR #167, opened 2025-08-01
chore: add engines field and bump node on CI

sjwilczynski:
In scope of this PR we bump node version on CI (from 18.x to 20.x as 18 is already after EOL and both 1JS and TMP are already on 22 and 20 respectively) and also add engines field to package json to inform devs about it.


alloy (approved):
🙏

PR #169, opened 2025-08-07
feat(nova/react-test-utils): make types for referenceEntries stronger

sjwilczynski:
We previously did some work to add good, strong types for parameters for nova decorator. One of the big things was making sure that `referenceEntries` fields match props of the component. However, we are now verifying only if the name of the prop matches, instead of of actually checking the value, which causes a number of random errors for folks using this in 1JS.

In scope of this PR we add a validation of return type of each field in `referenceEntries`. So previously

```
referenceEntries: {
       // the key "feedback" is checked against component props but the return value of the function is ignored
        feedback: (data) => data.feedback,
},
```

now

```
referenceEntries: {
       // also "data.feedback" is checked so that it matches the type that feedback prompt on the component expects
        feedback: (data) => data.feedback,
},
```

Thanks to @MikeBarker-MSFT for providing a first version of this within 1JS

### DIFF
--- a/change/@nova-examples-91d581e7-0d1e-4e1f-bafe-a9cead450b57.json
+++ b/change/@nova-examples-91d581e7-0d1e-4e1f-bafe-a9cead450b57.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "use stronger typing for reference entries",
+  "packageName": "@nova/examples",
+  "email": "stwilczy@microsoft.com",
+  "dependentChangeType": "patch"
+}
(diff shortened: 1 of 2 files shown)


packages/nova-react-test-utils/src/shared/storybook-nova-decorator-shared.tsx
```diff
@@ -32,15 +33,15 @@ export type UnknownOperation = {
 export type WithNovaEnvironment<
   TQuery extends OperationType = UnknownOperation,
   TypeMap extends DefaultMockResolvers = DefaultMockResolvers,
+  TComponent = never,
```
jakobkhansen, line 36: I'm not super well versed at generics in Typescript, but this is optional right? so it's backwards compatible with previous interface? If you don't supply it, it will be `never`?

MikeBarker-MSFT:
This looks like a REALLY good addition to the testing framework. Thank you for adding.

